PR #1, opened 2015-04-27
latest: Add cppcheck and doxygen for static tests

gebart:
Added cppcheck and Doxygen to use this image as a base system for CI testing. The static tests group requires that cppcheck and doxygen can be run in order to look for style errors.


OlegHahm:
ACK and go.

PR #3, opened 2015-05-21
Less layers by joining commands

gebart:
Each `RUN` command in the Dockerfile adds a file system layer to the container. Joining commands on a single line using `&&` results in less layers and less overhead.

This joins the update and dist-upgrade commands on a single line, there is really no use for them as separate layers.


OlegHahm:
I have no idea about docker, but the explanation sounds reasonable and the change seems valid. ACK and go!

PR #4, opened 2015-06-03
added GNU parallel

phiros:
The new drone build script uses GNU parallel to speed up the build process (See https://github.com/RIOT-OS/RIOT/pull/3038)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -59,6 +59,7 @@ RUN apt-get -y install \
     git \
     graphviz \
     libpcre3 \
+    parallel \
     pcregrep \
     python \
     python3 \


Dockerfile
```diff
@@ -65,7 +65,8 @@ RUN apt-get -y install \
     p7zip \
     subversion \
     unzip \
-    wget
+    wget \
+    parallel
```
gebart: nitpick: could you move `parallel` to above `pcregrep`?

gebart:
ACK, fix the nitpick if you want, otherwise merge at will.

phiros:
Ah didn't realize that this list was in alphabetical order -> fixed

PR #5, opened 2015-06-25
gcc 4.9

gebart:
This updates the gcc-arm-none-eabi version in the Docker image to 4.9 via the jessie-backports repository.

Wait until libstdc++-arm-none-eabi-newlib_4.9.2-21+7 hits the Debian package repository before merging.

see:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=787913
https://packages.debian.org/ca/jessie-backports/libstdc++-arm-none-eabi-newlib


OlegHahm:
ACK -. merge as soon as the Debian repository is updated accordingly.

PR #8, opened 2015-07-13
arm+native: Add clang

gebart:
The llvm package does not contain the main clang binary on Debian.

The original intention of the llvm package in the Dockerfiles was to provide clang inside the containers.


phiros:
Looks good -> ACK

PR #10, opened 2015-11-16
docker: add global system-wide git user and email address

cgundogan:
This PR sets up a global user and email address for git.
Rationale: When using docker to build a test that requires a package with patches, then
the patching returns the following error:
```
fatal: unable to look up current user in the passwd file: no such user
You need to set your committer info first
```
Can be invoked by issueing
```
BUILD_IN_DOCKER=1 BOARD=native make distclean clean all
```
in the `$(RIOTBASE)/tests/coap` folder.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -102,5 +102,9 @@ RUN apt-get -y install \
 # Create working directory for mounting the RIOT sources
 RUN mkdir -p /data/riotbuild
 
+# Set a global system-wide git user and email address
+RUN git config --system user.name "riot" && \
+    git config --system user.email "riot@example.com"
+
 WORKDIR /data/riotbuild
 


Dockerfile
```diff
@@ -102,5 +102,9 @@ RUN apt-get -y install \
 # Create working directory for mounting the RIOT sources
 RUN mkdir -p /data/riotbuild
 
+# Set a global git user and email address
+RUN git config --global user.email "riot@riot-os.org"
+RUN git config --global user.name "RIOT-Docker"
```
gebart: If you join the two lines with an `&&` we reduce the number of extra file system layers.

cgundogan:
@gebart nice hint. I amended the change

gebart:
ACK

OlegHahm:
Can we maybe use a different address?

OlegHahm:
I would prefer an address that does not forward per default to various real persons here.

gebart:
ah, good point

cgundogan:
@OlegHahm addressed comment and amended

cgundogan:
Hmm, I think the problem is not solved 100% with this patch.
When I build this image locally and use it for `BUILD_IN_DOCKER=1`, then I still receive the 
```
fatal: unable to look up current user in the passwd file: no such user
```
The other error is gone however

@gebart any ideas?

cgundogan:
@gebart This is what works for me now. I had to add a user with uid 1000, because `docker run` is called with `-u $(id -u)` which is `1000` on most systems. The error was that there was no user setup within the docker environment so that git couldn't look it up in the passwd file. Just removing the `-u` parameter from the `docker run` command also didn't work, because all artifacts would then belong to root and cleaning them will not work without root permissions.

Honestly, I dislike this solution of mine, because we cannot be sure that this make target is invoked with a ui of 1000. But I have no further ideas regarding this. Any opinions @gebart @OlegHahm ?

gebart:
I guess that `getent passwd $(id -u)` does not work inside the docker container without creating the user beforehand. Perhaps it would be possible to create the user passwd-entry dynamically on container creation?

kaspar030:
Use fakeroot for everything within docker? I did that for openwrt-builds inside docker.

cgundogan:
thank you all for your input. I figured that when setting a system-wide configuration then git won't do this passwd check. `BUILD_IN_DOCKER=1 BOARD=native make distclean clean all` works as intended now with this patch.

cgundogan:
@gebart does your ACK uphold?

gebart:
It would be more clean to use `git config --system` in my opinion

cgundogan:
@gebart you are right it is much cleaner this way. I amended the change immediately.

gebart:
ACK

cgundogan:
and GO

PR #11, opened 2016-01-28
Add signal wrapper

gebart:
Sometimes when pressing ctrl-c to abort a build the container doesn't exit and one make process usually starts to use 100% cpu in an infinite loop. This seem to be caused by not handling signals properly. This script adds a signal handler for SIGTERM (kill) and SIGINT (ctrl-c) to ask the subprocesses to terminate before exiting.

Tested on my own machine, both ctrl-c and kill, both directed at the top level sh process and the make and the gcc processes. I have not had a single case of stuck processes so far.

Bonus commit: Changed my last name where I could find it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,7 +32,7 @@ FROM debian:jessie
 # probably not a good idea to use this for your build needs.
 #FROM debian:experimental
 
-MAINTAINER Joakim Gebart <joakim.gebart@eistec.se>
+MAINTAINER Joakim Nohlgård <joakim.nohlgard@eistec.se>
 
 ENV DEBIAN_FRONTEND noninteractive
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -106,5 +106,8 @@ RUN mkdir -p /data/riotbuild
 RUN git config --system user.name "riot" && \
     git config --system user.email "riot@example.com"
 
-WORKDIR /data/riotbuild
+# Copy our entry point script (signal wrapper)
+COPY run.sh /run.sh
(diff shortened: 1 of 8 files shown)


gebart:
ping @cgundogan

kaspar030:
I think this would eat the exit status of the command being run.

kaspar030:
Would this force execution through run.sh when I do "docker -t -i <cmd>"?

If yes, I'd prefer to supply run.sh with the container, but not force it's usage by using it as ENTRY_POINT.

I'm using this image for the new CI system, and it took a while to make it both end correctly on SIGTERM and return the error code.

gebart:
Updated to return the exit code of the command to the docker process.
The signal handler is supposed to run as PID 1 in order to catch all signals from outside the container and it is supposed to be transparent to the user, so that we don't have to prefix the commands with `run.sh`.
Because this container is meant to be used by developers for building things, I think it makes sense to transparently add the signal handler as a wrapper to allow the users to run it as `docker run -ti --volumes-from blah -w /data/riotproject/examples/default riot/riotbuild make` and everything working as if you had run make without a container (CTRL+C works etc).

kaspar030:
Makes sense, and also run.sh doesn't interfere with the signal handling I built on top.

ACK.

kaspar030:
I just realized that before you could get a shell inside the container using "docker run -t -i riotdocker" (without command), which doesn't work anymore. Should we add handling for that?

gebart:
Good catch, we should fix that

gebart:
Opened #14 to fix the problem, run a default shell if no command is specified

PR #12, opened 2016-03-04
add cmake to Dockerfile

kaspar030:
relic needs it...


gebart:
ACK

kaspar030:
And go!

kaspar030:
@gebart How does an image here end up as being the one that gets pulled if I do "docker pull riot/riotbuild"?

gebart:
There's a hook to the Docker hub so it gets rebuilt automatically. While you're working with the containers, could you take a look at #11?

Issue #9, opened 2015-07-21
arm: Bump newlib to 2.2.0.20150623 to fix compile errors when compiling with clang


gebart:
Similar to what was seen on qemu-i386 with gcc.

See https://sourceware.org/ml/newlib/2015/msg00405.html
```
make: Entering directory '/data/riotbuild/riotproject/examples/ng_networking'
Building application "ng_networking" for "mulle" with MCU "k60".

In file included from /data/riotbuild/riotproject/examples/ng_networking/udp.c:26:
In file included from /data/riotbuild/riotproject/sys/include/net/ng_ipv6.h:38:
In file included from /data/riotbuild/riotproject/sys/include/net/ng_ipv6/nc.h:31:
In file included from /data/riotbuild/riotproject/sys/include/vtimer.h:28:
/data/riotbuild/riotproject/sys/include/timex.h:185:14: error: format specifies type 'unsigned long' but the argument has type
      'uint32_t' (aka 'unsigned int') [-Werror,-Wformat]
             t.seconds, t.microseconds);
             ^~~~~~~~~
/data/riotbuild/riotproject/sys/include/timex.h:185:25: error: format specifies type 'unsigned long' but the argument has type
      'uint32_t' (aka 'unsigned int') [-Werror,-Wformat]
             t.seconds, t.microseconds);
                        ^~~~~~~~~~~~~~
/data/riotbuild/riotproject/examples/ng_networking/udp.c:99:17: error: format specifies type 'unsigned long' but the argument has
      type 'uint32_t' (aka 'unsigned int') [-Werror,-Wformat]
                server.demux_ctx);
                ^~~~~~~~~~~~~~~~
3 errors generated.
```

gebart:
fixed by #15

PR #15, opened 2016-03-08
Use Ubuntu 15.10 Wily as base, switch to gcc-arm-embedded toolchain

gebart:
Updated Dockerfile to use Ubuntu 15.10 "Wily" as base image instead of Debian Jessie.
Also switched to gcc-arm-embedded toolchain (officially supported by ARM) which fixes the issue with `PRIu32` and Clang reported in #9 

Based on #14

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -112,4 +112,7 @@ RUN git config --system user.name "riot" && \
 COPY run.sh /run.sh
 ENTRYPOINT ["/bin/bash", "/run.sh"]
 
+# By default, run a shell when no command is specified on the docker command line
+CMD ["/bin/bash"]
+
 WORKDIR /data/riotbuild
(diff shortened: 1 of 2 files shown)


authmillenon:
Why not wait a few weeks for the 16.04 release. This will be LTS again.

gebart:
@authmillenon we will have to wait until the gcc-arm-embedded PPA provides xenial builds. Everything else is already working with 16.04 if you `s/wily/xenial/g` in this PR, but I decided that since it's such a simple change to upgrade it later we could just as well merge this now and move ahead to 16.04 when it's out.

kaspar030:
+1
I'll test with the CI later today.

kaspar030:
ACK & go.

gebart:
16.04 update is found in #21

PR #17, opened 2016-03-08
Clean up apt after installation

gebart:
based on #15 
This merges all apt commands into a single RUN line to avoid creating extra layers and removes the apt repository cache after installation finishes.
The image size is reduced by ca 30 MB. It's not much reduction for an image of 1.5 GB, but less layers can also slightly reduce the runtime overhead, which can be helpful when using lots and lots of ephemeral containers like for a CI build bot (RIOT-CI).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -112,4 +112,7 @@ RUN git config --system user.name "riot" && \
 COPY run.sh /run.sh
 ENTRYPOINT ["/bin/bash", "/run.sh"]
 
+# By default, run a shell when no command is specified on the docker command line
+CMD ["/bin/bash"]
+
 WORKDIR /data/riotbuild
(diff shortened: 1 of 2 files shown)


kaspar030:
ACK for the change, will test together with #15.

gebart:
@kaspar030 How did the test run go?

kaspar030:
@gebart sorry, didn't come further than buliding the docker image. will test now.

kaspar030:
Seems fine. ACK & go.

kaspar030:
no, I'll merge in order.

kaspar030:
Does this need a rebase now?

gebart:
Should be okay if github says it's OK

kaspar030:
Lets see :)

PR #18, opened 2016-03-11
Remove platform specific images

### LINKED ISSUE #16: Drop all platform-specific toolchains to reduce support effort?
Currently the `latest` label in Docker has support for all RIOT platforms, while the `arm`, `msp430` and `avr` labels only have support for those particular platforms. The original intent was to reduce the download size for people only interested in a specific platform, but I realize that it is a hassle to keep the Dockerfiles in sync.

Does anyone use the reduced functionality images or is everyone who is using docker to build RIOT using the full riot/riotbuild:latest image?

gebart:
Fixes #16


gebart:
I deleted the automatic triggers for building the other tags as well, now only the main Dockerfile is built (`latest` tag)

kaspar030:
ACK.

PR #20, opened 2016-03-23
dockerfile: add native gdb

kaspar030:
static-tests are probably soon running "make term" through gdb in order to see backtraces.


kaspar030:
(murdock is already running with this patch)

OlegHahm:
ACK

PR #22, opened 2016-08-11
add tool to create a passwd entry for user running the container

kaspar030:
This PR makes sure that the uid executing a command within the container has a proper passwd entry.

This solves some trouble with stuff requiring a user entry:

```
I have no name!@984e0102d179:/data/riotbuild$ ssh booze
No user exists for uid 1000
I have no name!@984e0102d179:/data/riotbuild$
```

vs.

```
riotbuild@bb9627754a88:~$ ssh booze
The authenticity of host 'booze (192.168.1.150)' can't be established.
ECDSA key fingerprint is SHA256:JgfpYCuH13M74r4kJnRt+yegIdQq5F5BW68FADcYf1I.
Are you sure you want to continue connecting (yes/no)?
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -84,6 +84,12 @@ RUN \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+# compile suid create_user binary
+COPY create_user.c /tmp/create_user.c
+RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.c -o /usr/local/bin/create_user \
+    && chown root:root /usr/local/bin/create_user \
+    && chmod u=rws,g=x,o=- /usr/local/bin/create_user
+
 # Create working directory for mounting the RIOT sources
 RUN mkdir -p /data/riotbuild
 
--- a/create_user.c
+++ b/create_user.c
@@ -1,0 +1,20 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+int main(int argc, char *argv[])
+{
+    if (argc < 2) {
+        return 1;
+    }
+
+    setuid(0);
+
+    unsigned uid = atoi(argv[1]);
+    char buf[128];
+
+    sprintf(buf, "/usr/sbin/useradd -u %u -d %s -r -g 0 -N %s", uid, HOMEDIR, USERNAME);
+    system(buf);
+    return 0;
+}
--- a/run.sh
+++ b/run.sh
@@ -26,6 +26,10 @@ runcommand() {
     return "$retval"
 }
 
+# create passwd entry for current uid, fix HOME variable
+create_user $(id -u)
+export HOME=/data/riotbuild
+
 if [ $# = 0 ]; then
     echo "$0: No command specified" >&2
     # docker run also exits with error code 125 when no command is specified and


gebart:
It seems a bit convoluted to make a C program for simply calling `useradd`. Can't it be called directly from the Dockerfile or run.sh instead?

kaspar030:
> It seems a bit convoluted to make a C program for simply calling useradd. Can't it be called directly from the Dockerfile or run.sh instead?

It can't be called from the dockerfile as the uid is not known at container build time.

This is using a C binary because useradd needs root permissions. Unless we want to suid root useradd, using a binary is the only way, as it is not possible to run shell scripts suid.

Dockerfile
```diff
@@ -84,6 +84,12 @@ RUN \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+# compile suid create_user binary
+COPY create_user.c /tmp/create_user.c
+RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.c -o /usr/local/bin/create_user \
+    && chown root:root /usr/local/bin/create_user \
+    && chmod 4750 /usr/local/bin/create_user
```
gebart: Does this actually work with no exec bit on others?

Could you change this to symbolic permissions? It makes it a bit easier to understand IMO.

cgundogan:
Generally, I use
```
ENV SOMEUSER test
ENV SOMEPASS 123
RUN useradd $SOMEUSER -m -p `openssl passwd -1 $SOMEPASS` -s /bin/zsh
```
in my docker files to have a correct user setup with a passwd entry. Would this also work for you?

Dockerfile
kaspar030: >> +# compile suid create_user binary
>> +    && chmod 4750 /usr/local/bin/create_user
> 
> Does this actually work with no exec bit on others?
Yes, as the default uid for "docker run" is 0, so the active user
belongs to the correct group.

> Could you change this to symbolic permissions? It makes it a bit easier to understand IMO.

Really? :) I never know how to specify the more complex ones using
symbolic permission...

kaspar030:
> Generally, I use
> ```
> ENV SOMEUSER test
> ENV SOMEPASS 123
> RUN useradd $SOMEUSER -m -p `openssl passwd -1 $SOMEPASS` -s /bin/zsh
> ```
> in my docker files to have a correct user setup with a passwd entry. Would this also work for you?

No.

The point is to be able to do "docker run -u $(id -u))", which might be
any user depending on the environment the container is launched. This
cannot be done within the Dockerfile.

It is possible to force a specific uid when launchin the container
(e.g., "docker run -u $uid_from_dockerfile"), but then the in-container
user cannot access files that belong to the user on the host system
(without the need to create the exact same user on the host system and
making sure it has access).

Dockerfile
kaspar030: > > Could you change this to symbolic permissions? It makes it a bit easier to understand IMO.
> Really? :) I never know how to specify the more complex ones using
> symbolic permission...

Or better, what's "4750" in symbolic notation?

gebart: u+rwxsg+x should do it I believe. The "s" is the suid bit

gebart: and group should only need execute permission, not read.

kaspar030: > and group should only need execute permission, not read.

Changed it to "u=rws,g=x,o=-".

cgundogan:
@kaspar030 we could also drop the `-u $(id -u)` parameter and let it run as root (or as any other user that we might create). To fix the ownership of the resulting binaries, we can modify `run.sh` to execute chown/chmod recursively once `run.sh` finishes (see https://github.com/BD2KGenomics/cgl-docker-lib/blob/master/mutect/runtime/wrapper.sh)

cgundogan:
>@kaspar030 we could also drop the -u $(id -u) parameter and let it run as root (or as any other user that we might create). To fix the ownership of the resulting binaries, we can modify run.sh to execute chown/chmod recursively once run.sh finishes (see https://github.com/BD2KGenomics/cgl-docker-lib/blob/master/mutect/runtime/wrapper.sh)

hm, this may not work perfectly.

did you read about the trick using gosu as outlined in https://denibertovic.com/posts/handling-permissions-with-docker-volumes/ ?

cgundogan:
we could also bake the correct user in while building the image ..
```
ARG USER_ID=0
RUN useradd dockeruser -u ${USER_ID}.....
....
USER ${USER_ID}
```
`docker build --build-arg USER_ID=$(id -u)`

I assume an image is handled on user-level (host system) anyways, so I don't think that an image baked for user1 (host system) can end up to be used with user2 (host system)

kaspar030:
> I assume an image is handled on user-level (host system) anyways, so I don't think that an image baked for user1 (host system) can end up to be used with user2 (host system)

Unfortunately not. "docker run riotbuild:latest" will pull a prebuilt
image from the docker registry.

kaspar030:
> we could also drop the `-u $(id -u)` parameter and let it run as root (or as any other user that we might create).

Not if we want to be able to build the sources of a host user, which get
mounted into the container as volume.

If anything inside the container is run as root, the host user will have
trouble with a lot of root-owned files.

If the build inside the container is run with a uid!=0, but not the same
as the host user, the container doesn't have write access to the host
user's files. Unless those are all world read/writable.

Do you feel it so clumsy to compile this little helper?

gebart:
I don't want to run anything as root in docker when using volumes. The risk of accidentally messing up is too big. I agree with Kaspar on this PR, after the explanation that a shell script can't be suid. Unfortunately we need to have a compiled program which creates a new user dynamically.

kaspar030:
> I agree with Kaspar on this PR, after the explanation that a shell script can't be suid. Unfortunately we need to have a compiled program which creates a new user dynamically. 

So is this ok to squash?

The tool itself is not new. I've been using it for our internal docker
build container at my Telekom project for a while now to work around all
the ssh issues, and it has proven to work there.

cgundogan:
Ok to squash from my side

Am 12.08.2016 09:50 schrieb "Kaspar Schleiser" <notifications@github.com>:

kaspar030:
- squashed

kaspar030:
Was there an ACK?

cgundogan:
ACK

kaspar030:
&go.

PR #23, opened 2016-11-06
dockerfile: add ccache

cgundogan:
Currently, Murdock is copying a [ccache](https://github.com/RIOT-OS/murdock-scripts) binary into the docker image at runtime.
@kaspar030 was there any reasoning to use a local binary instead of using ccache from upstream?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -44,6 +44,7 @@ RUN \
     apt-get -y install \
         bsdmainutils \
         build-essential \
+        ccache \
         cmake \
         curl \
         cppcheck \


kaspar030:
@cgundogan Yes, the last LTS didn't have the latest ccache version, so it lacked a feature Murdock needed. AFAIR it was per-cache-directory configuration files, not sure.

What version does the current Ubuntu install?

cgundogan:
@kaspar030 
```
root@0411764ecf4f:/data/riotbuild# ccache --version
ccache version 3.2.4
```
`vs`
```
~/murdock-scripts % ./ccache --version                                                                                                                                                    
ccache version 3.2.4
```

kaspar030:
looks good! ACK&go.

PR #26, opened 2017-01-27
misc improvements

kaspar030:
Here's a small collection of fixes for our Docker image:

1. cut off the cleaning step of the package installation phase

This is more a developing aid. Every time something needs to be added, a full re-installation of all packages is needed. With this PR, it becomes possible to just install packages after the main bunch, making use of docker's caching, saving a lot of time.

2. /tmp/create_user.c is leftover but unneeded

3. /data/riotbuild can not be written to if not overmounted using "docker run -v". There are use-cases where this causes trouble.


gebart:
Did you compare image sizes?
Moving the clean up to a separate step will introduce an additional layer in the image. I introduced the clean up to reduce the image size, but when it happens in a separate layer it won't reduce the image size, I believe.

kaspar030:
Sorry for the long silence. I dropped the first commit, for development I now usually just add an ```apt-get update``` just after the cleaning step, which is as good for saving time. :)

I've also changed the directory creation to use "-m 777" instead of an extra chmod call.

gebart:
please squash

kaspar030:
- squashed

Issue #27, opened 2017-03-19
Use docker for CI builds


Janfred:
In a university project I want to set up a CI environment for RIOT with rust support.
I've already changed the Dockerfile in my private project to support rust, but I'm now stuck at setting up the Gitlab-CI-Runner we use at our university.

The main problem is, that the ci-runner is running the commands as root. In your entrypoint run.sh you have a `create_user` which creates a user with the current username (where I'm not pretty sure why a user without a passwd entry can exist, but I haven't completely understood docker yet.)
In my case I have the problem, that the create_user either tells me `useradd: UID 0 is not unique` or (if I try execute `create_user` before) `useradd: user 'riotbuild' already exists` when I want to start the test script.

Do you have any ideas what I could do and explain to me why you need to create a user before?

I've already started some tests with my rust-addons, but I've always used the commandline start.

kaspar030:
> The main problem is, that the ci-runner is running the commands as root.

Can you change that? (Add "-u <other_id>" to 'docker run' command line...)

> In your entrypoint run.sh you have a |create_user| which creates a user
> with the current username (where I'm not pretty sure why a user without
> a passwd entry can exist, but I haven't completely understood docker yet.)

A process can be started with a uid regardless of it existing in passwd.
That's unrelated to Docker.

> In my case I have the problem, that the create_user either tells me
> |useradd: UID 0 is not unique| or (if I try execute |create_user|
> before) |useradd: user 'riotbuild' already exists| when I want to start
> the test script.
> 
> Do you have any ideas what I could do and explain to me why you need to
> create a user before?

If you absolutely cannot run your CI command with a differnt UID than 0,
you'll need to make the useradd command conditional (e.g., ```if [ "$(id
-u)" != 0 ]; then ...``` in run.sh.

The reason we're creating the user is that the docker container allows
building from a host system folder, but inside the container
(```BUILD_IN_DOCKER=1 make ..```). That means the needs to be able to
access the host user's files. In order not to have the files created by
the build process be owned by root, we execute make with the host user's
uid. Problem is that e.g., ssh and git fail if the calling user doesn't
exist. Our workaround is to just create the user entry in the
container's passwd.

(Also, there's a slight increase in security if the build process is not
run as root.)

Janfred:
> If you absolutely cannot run your CI command with a differnt UID than 0, you'll need to make the useradd command conditional (e.g., ```if [ "$(id -u)" != 0 ]; then ...``` in run.sh.

[...]

PR #28, opened 2017-03-25
Check if user already exists before creating user

Janfred:
This PR is implementing a solution for #27

It checks if a user already exists before executing the `create_user` script.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -29,7 +29,7 @@ runcommand() {
 # create passwd entry for current uid, fix HOME variable
 # only execute, if the current uid does not exist.
 if id $(id -u) >/dev/null 2>/dev/null; then
-    echo Not creating user with uid $(id -u) because he already exists
+    echo "Not creating user with uid $(id -u) because the user already exists"
 else
     create_user $(id -u)
 fi
--- a/run.sh
+++ b/run.sh
@@ -27,7 +27,12 @@ runcommand() {
 }
 
 # create passwd entry for current uid, fix HOME variable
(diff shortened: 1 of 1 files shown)


run.sh
```diff
@@ -27,7 +27,12 @@ runcommand() {
 }
 
 # create passwd entry for current uid, fix HOME variable
-create_user $(id -u)
+# only execute, if the current uid does not exist.
+if id $(id -u) >/dev/null 2>/dev/null; then
+    echo Not creating user with uid $(id -u) because he already exists
```
gebart: I don't want to be picky, but could you rewrite this message in a gender neutral form, please?

gebart: Also, please add double quotation marks around the message.

kaspar030:
- do we need the error message at all? (just asking)
- could you do "id -u" once at the beginning and put it in a variable? No need to call id more than once.

Janfred:
@kaspar030 The message was introduced by me for debugging, I thought it might be useful for production, but I can remove it.

The reason why I haven't put `id -u` in a variable is, that I test if `id -u` exits with an error return code. I could use $? but this is imho not good coding style and leads to difficult debugging process once another call is added before the `if`.

smlng:
please squash, otherwise I think its a good addition, and I ran into such an issue already. @kaspar030 you think its fine, too?

smlng:
oh wait, we can you GitHub merge/squash here

kaspar030:
> you think its fine, too?

Yes!

PR #29, opened 2017-04-20
add coccinelle to installed packages

kaspar030:
This PR adds installation of coccinelle, which I plan on integrating in the CI scripts.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,7 @@ RUN \
         build-essential \
         ccache \
         cmake \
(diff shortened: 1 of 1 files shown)


kaspar030:
> what is the reason to not add it next to the other static tools (native tool chain and build system functionality)?

No reason other than I didn't realize the other checkers are in the toolchain list. Fixed.

kaspar030:
> Fixed.

... and force pushed.

PR #30, opened 2017-05-14
Openocd support

erstrom:
Enable OpenOCD support in the docker container.

OpenOCD is built from source instead of using the binary package from the Ubuntu apt repo.
Rationale: Get an as up-to-date version of OpenOCD as possible.

I have tested the patches with an Arduino Zero board.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,6 +103,9 @@ RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" -DSHELL=\"/bin/ba
 # Create working directory for mounting the RIOT sources
 RUN mkdir -p /data/riotbuild
 
+# Change ownership of /data/riotbuild from root to default user
+RUN chown 1000:1000 /data/riotbuild
+
 # Set a global system-wide git user and email address
 RUN git config --system user.name "riot" && \
     git config --system user.email "riot@example.com"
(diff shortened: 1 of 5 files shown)


cgundogan:
@erstrom can you state the problem on why the latest openocd version is needed?

erstrom:
Binary packages provided by distros are often outdated.
For OpenOCD this means that newer dongles are not supported etc.
In my particular case, a few fixes/features have been added for CMSIS-DAP since version 0.9.0 of OpenOCD (0.9.0 is the current version in the Ubuntu Xenial repo).

Having the capability to build OpenOCD from source makes it easier to upgrade to newer versions in the future as well, since we don't need to rely on the apt repos to get updated.

smlng:
I'm against compiling packages from source without need, do we know of any issue with the OpenOCD version provided by Ubuntu 16.04 LTS and boards supported by RIOT? If not I suggest to postpone this, also note that we will likely update the base version of riotdocker to Ubuntu 18.04 LTS soon after its available - which brings openocd 0.10.0.

cgundogan:
@erstrom apart from the OpenOCD changes, the provided shell scripts `docker-run.sh` and `docker-start.sh` look promising. Care to propose them in a separate PR?

benpicco:
OpenOCD hasn't made a release in almost 4 years.
For many new chips it's necessary to compile it from source to being able to flash them.
Also RIOT support is only available on the development version.

So there would be value in having this in the riotdocker install.

maribu:
There has been no activity for quite some time - I think it is fine to close this. Anyone disagreeing can just re-open it.

(Also: I think it may be useful to build OpenOCD as package instead of using docker, as we do e.g. for EDBG. That would also allow applying the downstream flash drivers currently needed for GD32V and QN908x - which currently is a bit of a pain. It it avoids hassle to pass through devices into the container.)

PR #32, opened 2017-06-29
add murdock slave support

kaspar030:
This PR adds the changes necessary for our murdock slave containers. Previously I kept this in kaspar030/riotdocker, but it is becoming increasingly inconvenient to keep it updated, and it adds only little to the base image, and it is time to have changes go through the regular review process.

This would also be the first step to using dwq (murdocks distribution backend) as replacement for the current run-in-docker targets, and/or adding transparent in-cluster builds into our makefiles.

The PR only adds some packages and thus should not affect regular users (BUILD_IN_DOCKER=1 ...).
The image itself is currently run in production murdock.


kaspar030:
- rebased

kaspar030:
> I'm not convinced that it is good to mix Murdock stuff in here - though I know it makes perfect sense that Murdock uses the same docker image.

Can we maybe reconsider? Why do you think it is not good?

gebart:
@kaspar030 how about making a separate Murdock slave image which would be based on the riotbuild image?

gebart:
I don't like that this PR adds 3 more layers to our riotbuild image, making a separate Murdock slave image would have the same result without affecting the riotbuild image

kaspar030:
> how about making a separate Murdock slave image which would be based on the riotbuild image?

That sounds like a good idea. I'll try.

kaspar030:
> That sounds like a good idea. I'll try.

Done. Closing this PR.

PR #33, opened 2017-07-06
Dockerfile: add python3 crypto libraries

kYc0o:
Python3 crypto libraries are needed for Murdock to generate standard crypto keys in order to sign valid images.

For now this is only used for MCUBoot support.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,9 @@ RUN \
         python \
         python3 \
         python3-pexpect \
+        python3-crypto \
+        python3-pyasn1 \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -62,6 +62,9 @@ RUN \
         python \
         python3 \
         python3-pexpect \
+        python3-crypto \
+        python3-pyasn1 \
+        python3-ecdsa
```
gebart: Missing backslash

kYc0o: Done, directly amended.

PR #36, opened 2017-11-29
add flake8 dependency

aabadie:
Required for https://github.com/RIOT-OS/RIOT/pull/8078

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,7 +65,7 @@ RUN \
         python3-crypto \
         python3-pyasn1 \
         python3-ecdsa \
-        python3-pep8 \
+        python3-flake8 \
         p7zip \
         subversion \
         unzip \


aabadie:
I looked for one but didn't find. I'll double check because now I'm not sure anymore..

smlng:
python-flake8 from what I googled, but I guess its also fine to use pip, or not?

gebart:
I think pip is fine. I'm guessing that the package python-flake8 is for python 2.7, but I if it follows the naming conventions among the other Ubuntu packages.

kaspar030:
I've got an 16.04 box that offers both python-flake8 and python3-flake8.

The advantage of not using pip (apart from performance reasons) is that canonical takes care of the updating, while when using pip, the docker container becomes less reproducible.

IMO if there's a package, use it, so please change to use python3-flake8

aabadie:
Indeed python3-flake8 exists (just checked as well). I'll change that.

aabadie:
Done

kaspar030:
&go, I'll update the murdock container.

cgundogan:
>I'll update the murdock container.

wait a sec please

kaspar030:
> wait a sec please

ok. sth wrong?

cgundogan:
>ok. sth wrong?

Nope. I'm just about to provide a PR for cmake version >=3.7, because I need that for https://github.com/RIOT-OS/RIOT/issues/7713
I figured it's easier for you to update all murdock instances only once (:

cgundogan:
I meant https://github.com/RIOT-OS/RIOT/pull/7437

PR #38, opened 2017-11-30
cmake: install version 3.10

cgundogan:
In https://github.com/RIOT-OS/RIOT/pull/7437 we figured that our docker image is shipping a rather old CMake version. CCN-lite requires CMake >= 3.7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,7 +46,6 @@ RUN \
         bsdmainutils \
         build-essential \
         ccache \
-        cmake \
         coccinelle \
         curl \
         cppcheck \
@@ -95,6 +94,12 @@ RUN \
         libsocketcan2:i386 \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Install CMake 3.10
+RUN mkdir -p /opt && \
+    wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
+    | tar -C /opt -xz && cd /opt/cmake-3.10.0/ && ./bootstrap && \
+    make && make install && cd && rm -rf /opt/cmake-3.10.0
 
 # Install MIPS binary toolchain
 RUN mkdir -p /opt && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,10 +96,9 @@ RUN \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Install CMake 3.10
-RUN mkdir -p /opt && \
-    wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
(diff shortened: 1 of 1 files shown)


smlng:
how about updating to docker base image to `ubuntu:zesty` instead, which ships with cmake 3.7.2, see [here](https://packages.ubuntu.com/zesty/cmake).?

cgundogan:
I think switching to that's a pretty huge change.

gebart:
@smlng I think we had a discussion before that it was best to only use the LTS releases for the riotbuild image

cgundogan:
>Is there no pre-built package for Ubuntu?

not here https://cmake.org/download/

smlng:
@gebart docker says 17.04 (Zesty) is LTS, see [here](https://store.docker.com/editions/community/docker-ce-server-ubuntu)

smlng:
but ubuntu itself does not 🤔

cgundogan:
https://wiki.ubuntu.com/ZestyZapus/ReleaseNotes

miri64:
> @gebart docker says 17.04 (Zesty) is LTS, see here
>
> but ubuntu itself does not :thinking:

Ubuntu's policy on this is pretty clear (at least until the writing of this post): the April release of each even year is an LTS release (i.e. every 1X.04 release, where X is even ;-)).

Dockerfile
```diff
@@ -96,6 +95,12 @@ RUN \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+# Install CMake 3.10
+RUN mkdir -p /opt && \
+    wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
+    | tar -C /opt -xz && cd /opt/cmake-3.10.0/ && ./bootstrap && \
+    make && make install && cd && rm -rf /opt/cmake-3.10.0
```
smlng: why not use `/tmp` instead of `/opt`, as you delete it after installing anyway?

cgundogan: To be honest, I just copied the section from below. But I can use /tmp instead. Let me try if this still works and then I will adjust.

smlng: oh I oversaw that we use `/opt` already, mhm - either change both to `tmp` or stick with `opt`, but have it consistent, sorry for the noise.

cgundogan: I think having it in `/tmp` is nice, because it may be mapped to RAM in the host system. /opt however is supposed to be persistent for the mips toolchain.

cgundogan: @smlng the fixup commit changes that to /tmp, how do you feel about this?

smlng:
please squash

smlng:
or we use GitHub?

smlng:
squash and merge

cgundogan:
neat! @kaspar030 can you update the compile farm? (:

smlng:
> update the compile farm

can I haz #28 before?

smlng:
@cgundogan feel free to have a look at #28 too, I think I could use for the HAW ICC stuff.

PR #39, opened 2018-02-20
add risc-v toolchain

kaspar030:
This PR adds the RISC-V toolchain.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,6 +103,18 @@ RUN mkdir -p /opt && \
 ENV PATH $PATH:/opt/mips-mti-elf/2016.05-03/bin
 ENV MIPS_ELF_ROOT /opt/mips-mti-elf/2016.05-03
 
+# Install RISC-V binary toolchain
+RUN mkdir -p /opt && \
+        wget -q https://github.com/gnu-mcu-eclipse/riscv-none-gcc/releases/download/v7.2.0-2-20180110/gnu-mcu-eclipse-riscv-none-gcc-7.2.0-2-20180111-2230-centos64.tgz -O- \
+        | tar -C /opt -xz
+
+# HACK download arch linux' flex dynamic library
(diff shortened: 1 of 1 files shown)


kaspar030:
Got a problem:

```
riscv-none-embed-ar: error while loading shared libraries: libfl.so.2: cannot open shared object file: No such file or directory
```
... but ubuntu xenial's libfl is some black magic hack to libfl_pic.a ?

kaspar030:
> Got a problem:

Fixed it through injecting Arch Linux' libfl.so.2.0.0 ...

Dockerfile
```diff
@@ -103,6 +103,18 @@ RUN mkdir -p /opt && \
 ENV PATH $PATH:/opt/mips-mti-elf/2016.05-03/bin
 ENV MIPS_ELF_ROOT /opt/mips-mti-elf/2016.05-03
 
+# Install RISC-V binary toolchain
+RUN mkdir -p /opt && \
+        wget -q https://github.com/gnu-mcu-eclipse/riscv-none-gcc/releases/download/v7.2.0-2-20180110/gnu-mcu-eclipse-riscv-none-gcc-7.2.0-2-20180111-2230-centos64.tgz -O- \
+        | tar -C /opt -xz
+
+# HACK download arch linux' flex dynamic library
```
kaspar030: I know this is ugly... Hopefully bionic beaver will solve the issue.
Or are there other options?

gebart: Is there no RISC-V toolchain for Ubuntu?

kaspar030: You mean, pre-packaged? Didn't find one.
The eclipse one used in this PR claims "tested on Debian, Ubuntu, ...". :/

kaspar030:
This is blocking RISC-V-support. I propose going with the hack for now and revisiting when migrating to bionic beaver.

gebart:
Right, I don't mind

kaspar030:
Let's go.

kaspar030:
- squashed, waiting for travis

kaspar030:
&go

PR #40, opened 2018-03-27
dependencies: add automake

jia200x:
This PR adds `automake` to the RIOT docker image. It's (at least) required for building OpenThread.


aabadie:
> autoconf will be installed as a dependency of automake regardless

Indeed:

```
$ apt-cache rdepends automake | grep autoconf
autoconf
```

So I think we can merge this one.

PR #41, opened 2018-04-16
toolchain for ESP8266

gschorcht:
I have added build commands for ESP8266 toolchain as required for my EPS8266 port of RIOT, see PR [#8921](https://github.com/RIOT-OS/RIOT/pull/8921).

The toolchain is located in ```/opt/esp``` after installation and requires approximately 140 MB. Path variables in make point to this directory per default. 

I guess, adding ESP8266 toolchain to riotdocker is also a prerequisite for murdock test. Am I right?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -133,7 +133,7 @@ RUN echo 'Adding esp8266 toolchain' >&2 && \
     cd /opt && \
     git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-toolchain.git esp && \
     cd esp && \
-    git checkout v1.0
+    git checkout -q df38b06 
 
 ENV PATH $PATH:/opt/esp/esp-open-sdk/xtensa-lx106-elf/bin
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,12 +128,13 @@ RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.
     && chmod u=rws,g=x,o=- /usr/local/bin/create_user \
     && rm /tmp/create_user.c
 
-# Installs the complete ESP8266 toolchain in /opt/esp (170 MB after cleanup) from binaries
+# Installs the complete ESP8266 toolchain in /opt/esp (146 MB after cleanup) from binaries
 RUN echo 'Adding esp8266 toolchain' >&2 && \
     cd /opt && \
     git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-toolchain.git esp && \
     cd esp && \
-    git checkout -q df38b06 
+    git checkout -q df38b06 && \
+    rm -rf .git 
 
 ENV PATH $PATH:/opt/esp/esp-open-sdk/xtensa-lx106-elf/bin
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -128,6 +128,15 @@ RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.
     && chmod u=rws,g=x,o=- /usr/local/bin/create_user \
     && rm /tmp/create_user.c
 
+# Installs the complete ESP8266 toolchain in /opt/esp (170 MB after cleanup) from binaries
+RUN echo 'Adding esp8266 toolchain' >&2 && \
+    cd /opt && \
+    git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-toolchain.git esp && \
+    cd esp && \
+    git checkout v1.0
+
+ENV PATH $PATH:/opt/esp/esp-open-sdk/xtensa-lx106-elf/bin
+
 # Create working directory for mounting the RIOT sources
 RUN mkdir -m 777 -p /data/riotbuild
 


gebart:
Isn't the esp arm based?

gebart:
Sorry, I haven't been paying attention. You are correct in that Murdock does need this toolchain installed on the docker image to build the new port. But are there no precompiled packages available so that we don't need to compile newlib during the docker build?

gebart:
Not sure if the esp open sdk should go here or become a pkg in RIOT.

gschorcht:
No, ESP8266 isn't ARM-based. It uses Tensilica's configurable hardware cores called Xtensa.

esp-open-sdk is in fact not much more than crosstool-NG, the crosstool generator which builds the whole GCC tool suite for Xtensa platforms. It needs up to 25 minutes to compile and uses around 1.5 GB without cleaning (the default). From my understanding, packages are parts of RIOT that are given as source code and are compiled as part of RIOT. So, I'm not sure whether it would be possible at all to start the compilation of RIOT for ESP8266 boards without having the toolchain before.

Unfortunatly, there are no binary packages for eps-open-sdk and newlib. If you dont't think that it is a good idea to integrate the compilation process of the toolchain, I could try to provide it as binaries via a github repository.

gschorcht:
BTW, have seen that travis ci failed, propably due to timeout in the compilation process.

kYc0o:
> Unfortunatly, there are no binary packages for eps-open-sdk and newlib. If you dont't think that it is a good idea to integrate the compilation process of the toolchain, I could try to provide it as binaries via a github repository.

Yes, please. That will be the way to go, since for each PR Murdock is run and tests all examples for all platforms. Thus, it shouldn't build the toolchain each time Murdock is triggered, since we would take a lot of time for the builds (currently between 6 and 10 minutes).

kYc0o:
```
# Setup: (only needed once per Dockerfile change)
# 1. install docker, add yourself to docker group, enable docker, relogin
# 2. # docker build -t riotbuild .
```
Oops, I missed this part on the dockerfile. Indeed, it doesn't need a rebuild each time a PR is marked for CI tests. However, it's more convenient to only pull the binaries from somewhere, so we don't take too much time to re-deploy the image.

kYc0o:
> The job exceeded the maximum time limit for jobs, and has been terminated.

This is the error on Travis, so yes, it times out. So let's go for the binary solution if possible.

kYc0o:
>> Not sure if the esp open sdk should go here or become a pkg in RIOT.

> From my understanding, packages are parts of RIOT that are given as source code and are compiled as part of RIOT. So, I'm not sure whether it would be possible at all to start the compilation of RIOT for ESP8266 boards without having the toolchain before.

What it means is that the SDK can come separately, since it's not required directly by Murdock but by the code using it (Murdock executes the compile tests script which is essentially a "normal" compilation using `make all`).

In summary, the toolchain goes to murdock and the SDK as a RIOT package (in the RIOT repository).

gschorcht:
Just for clarification of myself. There are 3 components that are required:

- esp-open-sdk - even if it is called sdk, it is the toolchain in source code
- newlib - the standard libraries in source code
- ESP8266_NONOS_SDK - the vendor specific SDK as binary code

Since all three components are required to be able to get RIOT compiled, all of them belong to the tool chain from my point of view. The easiest way for me would be just to upload binary versions add a download and install command to the Dockerfile.

So, would it be problem to do it in that way?

gschorcht:
I think it wouldn't be a good idea to provide **esp-open-sdk** as a RIOT package. As I understand, packages are cloned and compiled for each application and each board in separate bin directories. Even though **esp-open-sdk** is called SDK, it is more like a complete tool suite and includes the whole **GCC** tool suite generated with **crosstool-NG** from source code. Generating this tool suite takes up to 25 min and needs 1.5 GByte. That would mean that each application and each board takes this compile time and the space. Am I wrong?

The SDK itself (the real SDK), comes as binary blobs from Espressif and can be optionally stored in a sub directory of **esp-open-sdk**.

Other components that are integrated in **esp-open-sdk** are libraries like **lwIP** that are not needed with RIOT. **esp-open-sdk** usually includes also a **newdir** library, but an older version without atomics that can't be used with RIOT.

kYc0o:
You may take a look [here](https://github.com/RIOT-OS/RIOT/blob/master/cpu/efm32/Makefile.include) for EFM based boards, which makes use of the package [here](https://github.com/RIOT-OS/RIOT/tree/master/pkg/gecko_sdk).

The goal is basically to download and compile the sources of the SDK, which contains mainly drivers for peripherals and initialisation code.

gschorcht:
I prepared a github repository with the binary version of the toolchain and changed the Dockerfile accordingly. Now, travis check passed and the toolchain seems to work in the docker image.

gschorcht:
@kYc0o Hi Francisco, what about this pull request to get the toolchain in riotdocker?

kYc0o:
Oh, I didn't know it wasn't merged. How is it possible then that your PR for the boards passed Murdock?

gschorcht:
@kYc0o The PR for the boards doesn't pass Murdock til now :-(

gschorcht:
Is there some progress with my pull request?

Dockerfile
```diff
@@ -127,6 +127,13 @@ RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.
     && chmod u=rws,g=x,o=- /usr/local/bin/create_user \
     && rm /tmp/create_user.c
 
+# Installs the complete ESP8266 toolchain in /opt/esp (170 MB after cleanup) from binaries
+RUN echo 'Adding esp8266 toolchain' >&2 && \
+    cd /opt && \
+    git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-toolchain.git esp
```
gebart: Do we want to use a particular tag or commit or do we just want the latest master commit?

kaspar030: yes, this needs to be pinned.

kYc0o: @gschorcht can you address the comments from @gebart and @kaspar030 ?

gschorcht: @gebart Oh, your question addressed my toolchain repository and not your riotdocker repository. Sorry, I did not realize it. Why not to use simply the last master commit? I don't think that it will change in future. If it does, I could change the docker file then and place a new PR.

gebart: You need to do it in two steps, first clone, then checkout.

gschorcht:
@gebart @kaspar030 Is that the pinning you meant?

gschorcht:
@kYc0o Your question is not really clear to me. Do you mean whether it is neccessary to execute the following command?
```
docker build -t riotbuild .
```
If so, then yes. Whenever the docker file changes, docker image has to be regenerated as for all other tool changes.

Dockerfile
```diff
@@ -128,6 +128,15 @@ RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.
     && chmod u=rws,g=x,o=- /usr/local/bin/create_user \
     && rm /tmp/create_user.c
 
+# Installs the complete ESP8266 toolchain in /opt/esp (170 MB after cleanup) from binaries
+RUN echo 'Adding esp8266 toolchain' >&2 && \
+    cd /opt && \
+    git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-toolchain.git esp && \
+    cd esp && \
+    git checkout v1.0
```
kaspar030: no, please use the actual commit hash.

kYc0o: True, a tag can be replaced at any time, I thought about it afterwards. It was a early ACK.

gschorcht:
Sorry, I'm a CVS and subversioner, git is completely new and different for me. Is it now ok?

kYc0o:
I'd say yes. @kaspar030 care to ACK?

Dockerfile
```diff
@@ -128,6 +128,15 @@ RUN gcc -DHOMEDIR=\"/data/riotbuild\" -DUSERNAME=\"riotbuild\" /tmp/create_user.
     && chmod u=rws,g=x,o=- /usr/local/bin/create_user \
     && rm /tmp/create_user.c
 
+# Installs the complete ESP8266 toolchain in /opt/esp (170 MB after cleanup) from binaries
+RUN echo 'Adding esp8266 toolchain' >&2 && \
+    cd /opt && \
+    git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-toolchain.git esp && \
+    cd esp && \
+    git checkout -q df38b06
```
kaspar030: can save 36M here by removing the .git folder (```&& rm -r .git```)

gschorcht: @kaspar030 Ok, done.

kYc0o:
Can we move forward on this? Looks ready and I'd really like to have ESP8266 into the release.

PR #44, opened 2018-07-29
esp8266: toolchain changes

gschorcht:
When I was trying not only to build ESP8266 images with riotdocker, but also to flash them, I realized that there are two additional requirements:

1. The python module for serial port access has to installed
2. The access to host's USB serial ports has to be permitted.

While requirement 1 is probabely specific for the ESP8266 toolchain, requirement 2 might also has to be fulfilled for other toolchains.

I made the following changes:

- to fulfill requirement 1, package ```python-serial``` is installed
- to fulfill requirement 2, user ```riotbuild``` is added to group ```dialout```

Starting the container by command
```
docker run -i -t --privileged -v /dev:/dev -u $UID -v $(pwd):/data/riotbuild riotbuild
```
then permits the access to host's USB serial devices, e.g., ```/dev/ttyUSB0```. Thus, riotdocker ca also be used to flash the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -66,6 +66,7 @@ RUN \
         python3-pyasn1 \
         python3-ecdsa \
         python3-flake8 \
+        python-serial \
         p7zip \
         subversion \
         unzip \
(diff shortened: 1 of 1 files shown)


gschorcht:
The changes are no longer needed. The python-serial package is added with the toolchain for the ESP32. Adding the user riotbuild to the group dialout is removed.

Issue #46, opened 2018-08-29
MIPS link is broken


jia200x:
I'm trying to rebuild the Docker image and just realized this MIPS package is broken:

http://codescape-mips-sdk.imgtec.com/components/toolchain/2016.05-03/Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz

That fails the image build

gebart:
See if this works better: http://codescape.mips.com/components/toolchain/2017.10-07/Codescape.GNU.Tools.Package.2017.10-07.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz

jia200x:
yes, that does the trick. I will psuh a fix. Thanks

cladmi:
I am testing a full rebuild now that #46 is merged and will close if it succeeds.

cladmi:
I successfully built the `master` branch so I will close this issue.

FlorianHeigl:
in case you see this, do you happen to still have this SDK someplace? I'm trying to archive all the MIPS SDK stuff, was mostly after Cavium but just found this is also important.

I can track that there were at least those two:
- Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz
- Codescape.GNU.Tools.Package.2019.09-01.for.MIPS.MTI.Bare.Metal.CentOS-6.x86_64.tar.gz

crasbe:
We *just* removed the the MIPS Toolchain the other day, but the link in the Dockerfile was still working: https://github.com/RIOT-OS/riotdocker/pull/263

Perhaps you can use it to find your versions?

FlorianHeigl:
Hi! So cool you replied!

I was lucky with the downloads, just came to say "found it". I'll leave a list here if it's ok, just for the next person.
Cannot praise Imagination enough for keeping the download links functional even when the main pages were removed. That is a good stewardship of products, it literally just takes one manager who cares to keep this online over website relaunches and they have one. exemplary.


Links I found:

- https://cloud.maschath.de/s/dDrZN4DSgJ33E3w/download/Codescape.GNU.Tools.Package.2019.09-03-2.for.MIPS32.MTI.Bare.Metal.Ubuntu-18.04.5.x86_64.tar.gz
[...]

PR #47, opened 2018-08-29
mips: Upgrade toolchain to version 2017.10-08

jia200x:
This PR fixes the MIPS toolchain link, as stated in #46 

### Issues/PRs references ###

Tracking the needed changes to update the toolchain: https://github.com/RIOT-OS/RIOT/pull/10009

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,7 +103,7 @@ RUN wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
 
 # Install MIPS binary toolchain
 RUN mkdir -p /opt && \
-        wget -q http://codescape-mips-sdk.imgtec.com/components/toolchain/2016.05-03/Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz -O- \
+        wget -q http://codescape.mips.com/components/toolchain/2017.10-07/Codescape.GNU.Tools.Package.2017.10-07.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz -O- \
         | tar -C /opt -xz
 
 ENV PATH $PATH:/opt/mips-mti-elf/2016.05-03/bin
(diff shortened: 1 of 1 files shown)


kaspar030:
Can someone test if the toolchain works as expected?

cladmi:
I will test.

Dockerfile
```diff
@@ -103,7 +103,7 @@ RUN wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
 
 # Install MIPS binary toolchain
 RUN mkdir -p /opt && \
-        wget -q http://codescape-mips-sdk.imgtec.com/components/toolchain/2016.05-03/Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz -O- \
+        wget -q http://codescape.mips.com/components/toolchain/2017.10-07/Codescape.GNU.Tools.Package.2017.10-07.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz -O- \
         | tar -C /opt -xz
 
 ENV PATH $PATH:/opt/mips-mti-elf/2016.05-03/bin
```
cladmi: This one also needs to be adapted `/opt/mips-img-elf/2017.10-08/bin`

cladmi:
FIrst the name changed from `mips-mti-elf-gcc` to `mips-img-elf-gcc`.

It also does not compile `examples/hello-world` with all 3 mips boards.

`error '_mips_handle_exception' redeclared with conflicting 'nomips16' attributes`

So there is some work to do.

cladmi:
If someone still has the old archive, it could  be hosted on the riot server until the support is updated.

kaspar030:
We do have the extracted toolchain in the deployed containers. While not preferable at all, we could tar.gz that up and use until someone finds the original file or we fix the compilation errors.

gebart:
I updated the PR with the URL to the latest release and updated the PATH variable.
See RIOT-OS/RIOT#10009 for a fix for the build issue

cladmi:
Found the issue I had while testing, I downloaded the 'IMG' archive instead of the 'MTI' one.

kaspar030:
Obsoleted by #76.

PR #52, opened 2018-10-05
Deduplicate binaries and remove toolchain documentation

gebart:
This saves around ~~70-80 MB~~ 130 MB in the finished image.

These toolchain binaries are hard links to the same file when installed from source. I think there was some mistake when the tar archives were made that failed to preserve the hard links and instead stores them as separate files. The toolchain tar archive for Cortex-M from ARM _does_ preserve the hard link property, so this file duplication has to be a mistake in the packaging for the RISC-V and MIPS toolchains.

Verified locally that these duplicate files have the same sha1 signature.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -104,7 +104,10 @@ RUN mkdir -p /opt && \
         wget -q https://codescape.mips.com/components/toolchain/2016.05-03/Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz -O- \
         | tar -C /opt -xz && \
     echo 'Removing documentation and translations' >&2 && \
-    rm -rf /opt/mips-mti-elf/*/share/{doc,info,man,locale}
+    rm -rf /opt/mips-mti-elf/*/share/{doc,info,man,locale} && \
+    echo 'Deduplicating binaries' >&2 && \
+    cd /opt/mips-mti-elf/*/mips-mti-elf/bin && \
+    for f in *; do rm "$f" && ln "../../bin/mips-mti-elf-$f" "$f"; done && cd -
 
 ENV PATH $PATH:/opt/mips-mti-elf/2016.05-03/bin
 ENV MIPS_ELF_ROOT /opt/mips-mti-elf/2016.05-03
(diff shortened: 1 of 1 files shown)


gebart:
Image sizes, latest is the docker hub latest, master was built locally from the master branch, 52 was built from this PR branch.
```
% docker images
REPOSITORY            TAG                 IMAGE ID            CREATED             SIZE
riot/riotbuild        52                  a812d4e35d42        About an hour ago   3.56GB
riot/riotbuild        master              6807a5d5eb9f        3 minutes ago       3.69GB
riot/riotbuild        latest              09f916d3a047        19 hours ago        3.69GB
```

cladmi:
I checked the archive and the deduplicated files have the same sha256.

I will try building it locally and build with it.

gebart:
@cladmi thanks!

cladmi:
I could build `examples/default` with both `pic32-wifire` and `esp8266-esp-12x`.

For the documentation, not sure anyone was using it from the container.
As it is a specific line that handles it, it can easily be deleted to rebuild with it enabled, either by a developer who needs it or later if we find it was an mistake.

For the locale, as we are not exporting `LC` I think other locales cannot be used anyway.

Dockerfile
```diff
@@ -100,15 +100,25 @@ RUN wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
 # Install MIPS binary toolchain
 RUN mkdir -p /opt && \
         wget -q https://codescape.mips.com/components/toolchain/2016.05-03/Codescape.GNU.Tools.Package.2016.05-03.for.MIPS.MTI.Bare.Metal.CentOS-5.x86_64.tar.gz -O- \
-        | tar -C /opt -xz
+        | tar -C /opt -xz && \
+    echo 'Removing documentation and translations' >&2 && \
+    rm -rf /opt/mips-mti-elf/*/share/{doc,info,man,locale} && \
+    echo 'Deduplicating binaries' && \
```
cladmi: Hmm noticed this line was not colorized when building but did not understand why. It is not printed to `stderr`.

Dockerfile
```diff
@@ -100,15 +100,25 @@ RUN wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
...
+ENV PATH ${PATH}:${MIPS_ELF_ROOT}/bin
 
 # Install RISC-V binary toolchain
 RUN mkdir -p /opt && \
         wget -q https://github.com/gnu-mcu-eclipse/riscv-none-gcc/releases/download/v7.2.0-2-20180110/gnu-mcu-eclipse-riscv-none-gcc-7.2.0-2-20180111-2230-centos64.tgz -O- \
-        | tar -C /opt -xz
+        | tar -C /opt -xz && \
+    echo 'Removing documentation' >&2 && \
+    rm -rf /opt/gnu-mcu-eclipse/riscv-none-gcc/*/share/doc && \
+    echo 'Deduplicating binaries' && \
```
cladmi: Same here, it is not printed to `stderr`.

gebart: ah, missed the redirection there. Will address

gebart:
@cladmi comments addressed, squashed

cladmi:
Monday was yesterday

PR #42, opened 2018-06-30
Update to Ubuntu Bionic

ZetaR60:
This updates to Ubuntu Bionic, which is the most current LTS release.

The ```team-gcc-arm-embedded``` PPA is removed, as it does not support Bionic. The package ```gcc-arm-embedded``` was from that PPA, but there is an equivalent ```gcc-arm-none-eabi``` that is probably newer anyway.

I tested with ```docker build -f Dockerfile ./``` and it builds successfully.

Required for: https://github.com/RIOT-OS/RIOT/issues/9405

PR #53, opened 2018-10-08
ARM: Use ARM provided toolchain archive instead of PPA

gebart:
The purpose of this change is to make the build more explicit with regards to toolchain versions, and to make the ARM installation more similar to the other supported platforms in the container. This will also make transitioning to a newer OS release (#42) easier.
The MD5 sum is verified after download against the expected MD5 sum from the ARM website, this way the user can be certain that this image uses a particular version of the toolchain.
The selected toolchain is the latest release currently available from arm.com, and the same release as found in the PPA. I did not verify that the files are the same in the downloaded archive as in the PPA, but they carry the same GCC version number, so the impact should be minimal.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,11 +33,6 @@ ENV DEBIAN_FRONTEND noninteractive
 # After adding the cleanup commands the size is approximately 1.497 GB
 RUN \
     dpkg --add-architecture i386 >&2 && \
-    echo 'Adding gcc-arm-embedded PPA' >&2 && \
-    echo "deb http://ppa.launchpad.net/team-gcc-arm-embedded/ppa/ubuntu xenial main" \
-     > /etc/apt/sources.list.d/gcc-arm-embedded.list && \
-    apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 \
-    --recv-keys B4D03348F75E3362B1E1C2A1D1FAA6ECF64D33B0 && \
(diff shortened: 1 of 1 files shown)


cladmi:
I am building the image and will check I can still do a "buildtest" with it.

(I will remove this `cmake` compiling from source it takes ages to compile and they already provide a binary version…).

cladmi:
We are currently already using this version (from the print_toolchain output in murdock).

    arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 7-2018-q3-update) 7.3.1 20180622 (release) [ARM/embedded-7-branch revision 261907]

Maybe we should directly use this one here.

gebart:
The latest available release on arm.com is 7-2018-q2-update. q3 is not available yet.

cladmi:
Ok.

cladmi:
Good for me, I could run `buildtest` for `examples/hello-world` and I checked it's using the `/opt/gcc-arm-none-eabi-7-2018-q2-update`.

I am completely in favor of fixing versions.

cladmi:
It is also 0.06GB smaller than `master`.

cladmi:
I rerun the build to be with the last master and have the travis output. Will merge on success.

cladmi:
Hmm… is not travis supposed to test the merge version with `master` ?
I do not see the build size output. Or maybe it is parsed before merging. Anyway ACK.

PR #55, opened 2018-10-10
Update to Ubuntu Bionic (rebased)

gebart:
Rebase of #42 because of inactivity.

- Dropped the compile from source cmake 3.10, it is already part of the upstream repos.
- Dropped arch linux libfl download

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -41,6 +41,7 @@ RUN \
         bsdmainutils \
         build-essential \
         ccache \
+        cmake \
         coccinelle \
         curl \
         cppcheck \
@@ -99,11 +100,6 @@ RUN echo 'Installing arm-none-eabi toolchain from arm.com' >&2 && \
     # No need to dedup, the ARM toolchain is already using hard links for the duplicated files
 
 ENV PATH ${PATH}:/opt/gcc-arm-none-eabi-7-2018-q2-update/bin
-
-# Install CMake 3.10
-RUN wget -q https://cmake.org/files/v3.10/cmake-3.10.0.tar.gz -O- \
-    | tar -C /tmp -xz && cd /tmp/cmake-3.10.0/ && ./bootstrap && \
-    make && make install && cd && rm -rf /tmp/cmake-3.10.0
 
 # Install MIPS binary toolchain
 RUN mkdir -p /opt && \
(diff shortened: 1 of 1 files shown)


smlng:
why not also use the arm-none-eabi-toolchain from the official repos, too? It's version 6.3.1, a bit older but could be worth trying to avoid too many external dependency that are likely to change.

kaspar030:
> I don't have any links here right low, but using the arm provided toolchain is a safer choice.

Wasn't there some issue with newlib compiled with large enums?

Anyhow, such a change should be independent of the underlying OS.

gebart:
Using the arm release also lets us make toolchain upgrades more explicit and visible. The OS release will matter less.

kb2ma:
I think #9248 is the reference about ARM toolchain issues with Bionic. Been there, done that.

smlng:
just tested with bionics arm-none-eabi toolchain, and sadly its still not working 😢 So no use in using the APT package ...

cladmi:
I would rather merge this just after the release than before.
I can take care of running the release build tests and some of the tests on this version.

cladmi:
But other than this I agree with the changes.

gebart:
@cladmi could you add a "request for changes" review so that we don't forget and merge this anyway?

I agree we should run a bunch of tests on this new version before merging. Especially the distro provided tools such as doxygen, cppcheck, coccinelle etc. but also the downloaded toolchains (RISC-V required the libfl.so hack in the older image, should be tested here)

kaspar030:
> I agree we should run a bunch of tests on this new version before merging.

How about we extend the travis script to do that.

e.g.,
- checkout riot master
- buildtest in examples/hello-world
- static test

cladmi:
> How about we extend the travis script to do that.
> 
> e.g.,
> 
>     * checkout riot master
> 
>     * buildtest in examples/hello-world
> 
>     * static test

Good idea, but it does not replace running extended tests, or we would not have murdock building everything and release tests.

cladmi:
@gebart I agree, that is why I said it is a good idea.
I was just precising I want more extended tests for this specific PR.

kYc0o:
@ZetaR60 can you confirm that this PR also fixes your time.h issue described on https://github.com/RIOT-OS/RIOT/pull/8842 ?

ZetaR60:
@kYc0o Yes. Updating to Bionic (and thus updating avr-libc) fixes the issues with time.h.

gebart:
This PR was a mistake by me, I don't know why I opened this in such a hurry.. Let's continue on #42 instead.

ZetaR60:
Thanks @gebart

PR #56, opened 2018-10-10
use python pip to install required packages

smlng:
This changes the install procedure for python packages from using APT packages to use python-pip and install packages from a file. This has several advantages:

- not all pythons deps for RIOT are available as APT packages (see Murdocks `dwq` for instance)
- installing from file makes dependencies more explicit
- and also allows to specify (and fix) package version (if needed)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,11 +57,7 @@ RUN \
         pcregrep \
         python \
         python3 \
-        python3-pexpect \
-        python3-crypto \
-        python3-pyasn1 \
-        python3-ecdsa \
-        python3-flake8 \
+        python3-pip \
         p7zip \
         subversion \
         unzip \
@@ -85,6 +81,12 @@ RUN \
         libsocketcan2:i386 \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# install required python packages from file
+COPY pydeps.txt /tmp/pydeps.txt
+RUN echo 'Installing python3 packages' >&2 \
+    && pip3 install -r /tmp/pydeps.txt \
+    && rm /tmp/pydeps.txt
 
 # Install ARM GNU embedded toolchain
 # For updates, see https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads
--- a/pydeps.txt
+++ b/pydeps.txt
@@ -1,0 +1,5 @@
+asn1
+ecdsa
+flake8
+pexpect
+pycrypto


kaspar030:
Hm, isn't pip a moving target? Using repo-packages brings at least some stability regarding versions.
I'd rather take what's in the repos if available. We can always override on a case-per-case bases.

smlng:
if necessary lets fix the versions in the `pydeps.txt` file, then. My point is for consistency, too - why install some packages via APT and other via PIP?

smlng:
@kaspar030 I fixed the versions for installation via `pip3` to match the ones in the xenial package repo.

pypkgs.txt
```diff
@@ -0,0 +1,6 @@
+pyasn1==0.1.9
```
aabadie: In general, python projects call this file `requirements.txt`. Moreoever  I wouldn't define it in riotdocker but create it directly somewhere in the RIOT codebase.

miri64: Some of those don't even have anything to do with RIOT directly (e.g. `flake8`) but are just used by CI for checks.

Dockerfile
```diff
@@ -95,6 +91,15 @@ RUN \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+# install required python packages from file
+COPY pydeps.txt /tmp/pydeps.txt
+COPY pypkgs.txt /tmp/pypkgs.txt
+RUN echo 'Installing python3 packages' >&2 \
+    && pip3 install -r /tmp/pydeps.txt \
```
kaspar030: why two files (pydeps and pypkgs)?

smlng: the pypkgs.txt contains all the actually wanted python packages like ASN1, i.e. these match the `apt-get install <python3-foo>`. The ones in `pydeps.txt` are prerequisites to be able to install those packages, i.e. they were not required when installing via `apt`.

smlng: ah and that way I can easily remove these packages in pydeps.txt after installing the required packages ... I'll add this to the Dockerfile, might safe some bytes

pypkgs.txt
aabadie: Well not really: flake8 could be used locally by a developer to do last time static checks before opening a PR.

kaspar030:
@gebart could you take a quick look?
@cladmi could you do your testing magic?

Dockerfile
```diff
@@ -95,6 +91,15 @@ RUN \
     && echo 'Cleaning up installation files' >&2 && \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
+# install required python packages from file
+COPY pydeps.txt /tmp/pydeps.txt
```
aabadie: Maybe it was unclear (or voluntary not taken in to account) but these files should be called <sthg>_requirements.txt. As stated in the [pip documentation](https://pip.readthedocs.io/en/latest/user_guide/#requirements-files).

Moreover, they should be defined in the RIOT repository and not the docker file. This will obviously help new developers setting-up their development environment!

smlng: no in practice you wouldn't build the container image yourself, you pull it from docker hub - which we support with this repo. See https://hub.docker.com/r/riot/riotbuild/

aabadie: > this allows for testing and publishing the container

This is also the case with a Dockerfile in RIOT by the way.

> makes reviewing such things independent of the RIOT reviewing process

What is wrong with the RIOT reviewing process ? too strict ? ok

smlng: > This is also the case with a Dockerfile in RIOT by the way.

how? Murdock won't try to build the Dockerfile and why should it? Here we have a distinct travis config that serves the task of building - maybe we should add some testing, too.

> What is wrong with the RIOT reviewing process ? too strict ? ok

no, but it's meant for code review ... not tooling and config - and with nearly 500 open PRs for mostly code stuff, things like this will likely get lost, delayed and forgotten.

---

divide and conquer, remember 😉

Dockerfile
```diff
@@ -92,13 +92,14 @@ RUN \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # install required python packages from file
-COPY pydeps.txt /tmp/pydeps.txt
-COPY pypkgs.txt /tmp/pypkgs.txt
+COPY deps_requirements.txt /tmp/deps_requirements.txt
```
kaspar030: please rename to something like "install_requirements.txt". "deps_requirements" seems like doubly the same as "requirements" which come right below.

miri64: `python_requirements.txt`? `install_requirements.txt` takes away the python context ;-)

smlng: @kaspar030 good suggestions!

Dockerfile
kaspar030: @aabadie so you do want to have this discussion here :)

Currently, there's some limited devops deployment strategy with the riotdocker repo. Every commit to master triggers a Docker hub rebuild, which when successful triggers a Murdock worker container rebuild, which if successful will then automatically be picked up by the workers eventually.

Docker hub rebuilds on push. We don't want that on every push to RIOT master.
Can we theoretically think of (script-fu or whatever) ways around that? Of course. Is it worth the hassle? I don't think so.

Having the Dockerfile in a seperate repository makes handling it much less complex. As long as there's no pressing reason to keep the Docker stuff in-sync with the main repository, that is just the sane way to go.

smlng:
rebased and squashed

miri64:
Apparently there is another advantage to using `pip`: https://github.com/RIOT-OS/riotdocker/network/alerts.  @smlng you think you can fix this?

smlng:
the link is not working for me, what is/was there to see/read?

miri64:
Don't you see the big banner at https://github.com/RIOT-OS/riotdocker/? It gives an overview about vulnerabilities in the imported python libs.

smlng:
nope big banner in my browser (Safari)

miri64:
image

image

image

miri64:
Mh... If I'm not logged in I don't see it either. Maybe you need certain permissions to see it.

smlng:
and according to the alert description there is currently no fix to it, so we need to wait. Btw. such issue show that we should check and rebuild the docker image more often to get latest security updates than just on changes to the package list

miri64:
(okay I was mistaken: 2017, but still)

PR #59, opened 2018-10-21
apt: Use --no-install-recommends to avoid unnecessary extras 

gebart:
Add `--no-install-recommends` to apt-get commands to avoid pulling in optional dependencies which are not used by the build process. 

The image size reduction is approximately 430 MB
```
REPOSITORY                 TAG                 IMAGE ID            CREATED             SIZE
riot/riotbuild             norecommends        402547be3514        3 minutes ago       3.34 GB
riot/riotbuild             master              e4c568fc283a        44 minutes ago      3.77 GB
```

## Testing

This needs a full buildtest run on the new image to ensure that no critical components are missing after adding this flag, but my laptop does not have the disk space or time to do it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,10 +36,11 @@ RUN \
     echo 'Upgrading all system packages to the latest available versions' >&2 && \
     apt-get update && apt-get -y dist-upgrade \
     && echo 'Installing native toolchain and build system functionality' >&2 && \
-    apt-get -y install \
+    apt-get -y --no-install-recommends install \
         automake \
         bsdmainutils \
         build-essential \
+        ca-certificates \
         ccache \
         coccinelle \
         curl \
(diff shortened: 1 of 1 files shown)


cladmi:
I will run the buildtest during the release with this version.

gebart:
Thanks, @cladmi!

cladmi:
I ran the compilation test with this, `scan-build` in one example, and also the `static-test`.

I will merge after the release.

cladmi:
Only failed build was tests/gnrc_ipv6_nib but it also fails with the current image.

`BOARD=waspmote-pro make -C tests/gnrc_ipv6_nib`

https://github.com/RIOT-OS/Release-Specs/issues/76#issuecomment-434270312

cladmi:
@smlng @kaspar030 do you see any other checks that should be done ? like for using it in murdock or something I missed

gebart:
Rebased

PR #61, opened 2018-11-18
Dockerfile: fix link to flex library

### LINKED ISSUE #60: Docker build is broken
Daily cron jobs are configured on Travis and they are reporting a build failure of the Docker image. See the output of the [last failure](https://travis-ci.org/RIOT-OS/riotdocker/builds/452898067).

There's something wrong with that line:
```
Step 12/26 : RUN wget -q https://sgp.mirror.pkgbuild.com/core/os/x86_64/flex-2.6.4-1-x86_64.pkg.tar.xz -O-         | tar -C / -xJ usr/lib/libfl.so.2.0.0
 ---> Running in d8fec91482a9
xz: (stdin): File format not recognized
tar: Child returned status 1
tar: Error is not recoverable: exiting now
The command '/bin/sh -c wget -q https://sgp.mirror.pkgbuild.com/core/os/x86_64/flex-2.6.4-1-x86_64.pkg.tar.xz -O-         | tar -C / -xJ usr/lib/libfl.so.2.0.0' returned a non-zero code: 2
```

aabadie:
fixes #60 by updating the link to the flex library.

I'm wondering why flex is required this way (the comment says it's a hack but doesn't say why).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -131,7 +131,7 @@ RUN mkdir -p /opt && \
     for f in *; do rm "$f" && ln "../../bin/riscv-none-embed-$f" "$f"; done && cd -
 
 # HACK download arch linux' flex dynamic library
-RUN wget -q https://sgp.mirror.pkgbuild.com/core/os/x86_64/flex-2.6.4-1-x86_64.pkg.tar.xz -O- \
+RUN wget -q https://sgp.mirror.pkgbuild.com/core/os/x86_64/flex-2.6.4-2-x86_64.pkg.tar.xz -O- \
         | tar -C / -xJ usr/lib/libfl.so.2.0.0
 RUN ldconfig
 


cladmi:
The reference PR adding the line say it 

https://github.com/RIOT-OS/riotdocker/pull/39#issuecomment-366956207
https://github.com/RIOT-OS/riotdocker/pull/39#issuecomment-366962286

cladmi:
BTW is it not a bad idea to rely on `arch-linux` repository as they basically only ever keep one version in the mirrors ?
This means no version can be build in the future.

At least relying on the archive would be better.

cladmi:
Not yet at least https://github.com/RIOT-OS/riotdocker/pull/42/files does not remove it and @gebart did not request to do it after he closed https://github.com/RIOT-OS/riotdocker/pull/55/files.

cladmi:
I also did not re-test, because, if it would have worked with the ubuntu bionic version, why using the `arch linux` version in the first place ?

PR #63, opened 2019-01-15
add package: lsb-release

jcarrano:
## Description

The static tests use `/etc/os-release` to report the OS version in Linux. This file is not standard: it is present in all systems with systemd and maybe other distros too. The lsb_release tool, on the other hand, is standardized as part of Linux Standard Base and should be the "correct" way to detect the OS version.

## Related PRs

Needed by https://github.com/RIOT-OS/RIOT/pull/10770

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,6 +55,7 @@ RUN \
         less \
         libpcre3 \
         libtool \
+        lsb-release \
         m4 \
         parallel \
         pcregrep \


gebart:
NACK, why add something to this image that we don't need?

kaspar030:
RIOT-OS/RIOT#10770 has been closed, so there's no need as @gebart suggested. Please reopen if this changes.

PR #65, opened 2019-01-18
esp32: add reminder to memset() problem

miri64:
In https://github.com/RIOT-OS/RIOT/pull/10801#pullrequestreview-194277653 I made a promise to remember that we check if `memset()` still isn't optimized out for sensitive data. This comment hopefully helps us to keep that promise.


miri64:
> Do you think that this is really necessary? The according files with `memset` calls are not used directly from original SDK. Instead a version adopted for RIOT is used.

This has nothing to do with the SDK but the optimizer. A toolchain update might change optimization behavior so while it might not disregard your efforts to not optimize it away in the current version, it might in others.

miri64:
(and this should be checked whenever the toolchain is updated)

miri64:
Done.

fjmolinas:
GO!

PR #67, opened 2019-02-11
add TI binary distribution of msp430-gcc

geromueller:


kaspar030:
I've updated this in #82.

kaspar030:
this is obsolete now

PR #68, opened 2019-02-27
create /etc/riotbuild_version with "git describe --always" output

kaspar030:
Currently, there's no way to get the git commit of the running docker container from within the container.

This PR uses Docker hub magic (namely, a custom build hook) to get the output of "git describe" into the container image.

See https://github.com/docker/hub-feedback/issues/600 for others having similar issue and solution.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -284,4 +284,19 @@ ENTRYPOINT ["/bin/bash", "/run.sh"]
 # By default, run a shell when no command is specified on the docker command line
 CMD ["/bin/bash"]
 
+# get Dockerfile version from build args
+ARG RIOTBUILD_VERSION=unknown
+ENV RIOTBUILD_VERSION $RIOTBUILD_VERSION
+
+ARG RIOTBUILD_COMMIT=unknown
+ENV RIOTBUILD_COMMIT $RIOTBUILD_COMMIT
+
+ARG RIOTBUILD_BRANCH=unknown
+ENV RIOTBUILD_BRANCH $RIOTBUILD_BRANCH
+
+# watch for single ">" vs double ">>"!
+RUN echo "RIOTBUILD_VERSION=$RIOTBUILD_VERSION" > /etc/riotbuild
+RUN echo "RIOTBUILD_COMMIT=$RIOTBUILD_COMMIT" >> /etc/riotbuild
+RUN echo "RIOTBUILD_BRANCH=$RIOTBUILD_BRANCH" >> /etc/riotbuild
+
 WORKDIR /data/riotbuild
--- a/hooks/build
+++ b/hooks/build
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+docker build \
+    --build-arg RIOTBUILD_BRANCH="$(git rev-parse --abbrev-ref HEAD)" \
+    --build-arg RIOTBUILD_COMMIT="$(git rev-parse HEAD)" \
+    --build-arg RIOTBUILD_VERSION="$(git describe --always)" \
+    -f "${DOCKERFILE_PATH}" -t "${IMAGE_NAME}" .


jcarrano:
I agree that it is a good idea to have the container self-identify. I'm not sure if the abbreviated commit hash is the best way:

- How are we going to use it?
- Why not the full commit hash?
- Why not the branch name?

kaspar030:
>  I'm not sure if the abbreviated commit hash is the best way:

As soon as we tag properly, it'll become something like "2019.04-devel-308-gf2e1d0dcb2"

>     * How are we going to use it?

At least, output it in every CI job run.

>     * Why not the full commit hash?

We're humans... So I prefer the "git describe" output. The hash is a fallback. It is also unique (at the point the git call is made). It doesn't really matter, we could even just use both.

>     * Why not the branch name?

Could be a good addition.

Maybe go for a shell-sourceable file:

```
ROITBUILD_GIT_BRANCH=master
RIOTBUILD_GIT_COMMIT=hexfoo
RIOTBUILD_GIT_DESCRIBE=production-2019.04-gfoo-dirty
```

kaspar030:
> ROITBUILD_GIT_BRANCH=master
> RIOTBUILD_GIT_COMMIT=hexfoo
> RIOTBUILD_GIT_DESCRIBE=production-2019.04-gfoo-dirty

I've implemented it as above, without the typo. @jcarrano mind to take another look?

kaspar030:
- rebased

fjmolinas:
@jcarrano ping!

jcarrano:
@cladmi you had some comments on this if I remember correctly?

cladmi:
It is just faking to have a correct versioning.

Also, branch name and git-describe are local specific information, so not really what should be part of a build.

kaspar030:
> It is just faking to have a correct versioning.

I think we're talking about different things. A container's "version" cannot be more exact than the git commit of its Dockerfile. (let's ignore for now that ubuntu:latest + "apt-get dist-upgrade" is a moving target).

That's all this PR adds, and only if the container is built with actually supplying the data via arguments to "docker build". Docker Hub would do that.

"Currently, there's no way to get the git commit of the running docker container from within the container."

This PR solves that problem. I need that so I can confirm a specific container's version.

> Also, branch name and git-describe are local specific information, so not really what should be part of a build.

branch name and "git describe" output might be local, if the container is built locally while supplying the correct arguments to docker build. So what?

Let's keep "versioning", if needed, seperate.

cladmi:
Yes for commit hash, I commented on the others.
However it is still internal, so you need to look into the container to know which one it is on your machine, and cannot request "build on the container that has /etc_riotbuild_version | grep hash"

If you have a version, that is visible outside, you can also write it inside.

So I still find it a hack for not doing versioning.

kaspar030:
> Yes for commit hash, I commented on the others.

I see. If I add "RIOTBUILD_BUILT_BY_DOCKER_HUB=0|1", would that help the locality issue?

To me the huge advantage of git-describe (in combination with meaningful tags) is that you see right away if something is younger or older than a specific commit.

E.g., is abbabbaba before bbabbafa? Is 2019.04-gabbabbaba before 2019.07-gbbabbafa?

> However it is still internal, so you need to look into the container to know which one it is on your machine, and cannot request "build on the container that has /etc_riotbuild_version | grep hash"

Well, we could tag the container with the commit hash. Or with the "git-describe" output. Or both.
In combination with only us having the possibility to tag on docker hub, this is pretty unique.
Does docker always retrieve all tags? Because none is gonna pull riot/riotbuild:12345. Some might pull the latest, others will pull a release tag version. Unless docker shows *all* tags of a certain image, in order to use *the exact same container*, we'd go back to container digests anyways.

> If you have a version, that is visible outside, you can also write it inside.
> 
> So I still find it a hack for not doing versioning.

I want to avoid having to maintain a self-managed version number within the container. Imagine the trouble when there are a couple of open pull requests:

1. current version is e.g., 12345
2. five open pull requests do minor changes, each updating /etc/version from 12345 to 12346
3. random PR gets merged
4. *all four other pull requests* have to increase the version number by one, rebasing, rebuilding,...
5. repeat at 2 (with one less PR or some other joining, current version is now 12346).

That's unmaintainable.

kaspar030:
Could someone take a look at this? It is discussion-blocked. I'm really missing a way of getting some kind of version from within the container.

kaspar030:
> If you have a version, that is visible outside, you can also write it inside.

I think it is more like the other way around. Once a container is built, it is immutable. Change it, and you have a new container.

Each container is uniquely identified by it's hash, but that is not available from within the container. (Are we the only ones that might use this information?)

If we manually version our containers, that can be done by tagging the resulting container's hash. with a version. That in turn needs to be done on the host that builds the container, which would be the Docker hub at the moment. Again, this information can not be accessed from within the container, unless we first tag, then build and pass the tag into the container.
This PR adds the infrastructure to pass a version into the container, at container build time.
Which version to put is up to the builder. This PR makes Docker hub use the commit hash, git describe output and branch name, but that can be changed.

> So I still find it a hack for not doing versioning.

This PR is a first step to any versioning that is visible from within the container.

fjmolinas:
> I want to avoid having to maintain a self-managed version number within the container. Imagine the trouble when there are a couple of open pull requests:

Why would this have to be self maintained? And why do it manually? Cant we handle that in `hooks/build`, it will of course also require passing the version so +1 to `RIOT_VERSION`. Using git describe for now seems OK to me. But we could have a `VERSION` which is bumped by the `hooks/build` scrip right?

Otherwise I think this is a good first step towards versioning.

kaspar030:
> Why would this have to be self maintained? And why do it manually?

well, we need somewhere to store that version. in travis the build has access to some build id which might be increasing and unique, but is a bit out of our control. and it doesn't really convey the info I need: from which git version was the container built?

kaspar030:
Can I get some 'maintainer wants it' slack here? :)

fjmolinas:
> Can I get some 'maintainer wants it' slack here? :)

If you want it so bad, why answer now a comment I made in March. This attitude really annoys me, I take time to look at your PR, I give you slack since I say "Otherwise I think this is a good first step towards versioning." (you can see I would have been OK with it), I look at what the others are saying and suggest something to which I get no answer.

So again,  why not do something like:

- https://medium.com/better-programming/how-to-version-your-docker-images-1d5c577ebf54

> well, we need somewhere to store that version. 

Why not a `VERSION` file in the repo?

kaspar030:
> Why not a `VERSION` file in the repo?

I thought that was answered: https://github.com/RIOT-OS/riotdocker/pull/68#issuecomment-522719705

Because every PR changes the "version". They don't necessarily get merged in order, so even if every PR would bump VERSION, it is a nightmare to keep it monotonic and in the right order. That's what I call "manually managing it". https://github.com/RIOT-OS/riotdocker/pull/68#issuecomment-522719705

The linked article makes the container build script modify the repository it is building from, in some "release.sh", to bump the version. This assumes that script is run by someone, manually. It is not meant to be used in the docker hub build hook. Could we run it there? Sure, with tons of logic to ensure this is only happening on merged builds, and ensuring that docker hub can actually push to our repository, and bypassing the branch protections. I don't see this as viable.


> If you want it so bad, why answer now a comment I made in March. This attitude really annoys me, I take time to look at your PR, I give you slack since I say "Otherwise I think this is a good first step towards versioning." (you can see I would have been OK with it), I look at what the others are saying and suggest something to which I get no answer.

Point taken for not answering. 

I don't want it so bad. The need for this pops up once every few months, where I manually go check container versions (through inference on installed pip versions, or by just updating to the newest). It is a complete side issue, but it is annoying at those times.

The "can I get some slack here" was meant as "yeah there might be other solutions with different tradeoffs, but unless you know a solution (*1) that is substantially better while solving the issue and not requiring not only starting from scratch with the implementation but requiring more work than the proposed solution, let's not spend more time on an exhaustive research on the solution space for such a simple requirement".

fjmolinas:
> Because every PR changes the "version". They don't necessarily get merged in order, so even if every PR would bump VERSION, it is a nightmare to keep it monotonic and in the right order. That's what I call "manually managing it". #68 (comment)

I though this could be done In the build-hook, I though I had tried it our when I made the comment in March, but maybe that is not the case since I can't fin any branch where I would have done it.

> I don't want it so bad. The need for this pops up once every few months, where I manually go check container versions (through inference on installed pip versions, or by just updating to the newest). It is a complete side issue, but it is annoying at those times.

Anyway I was fine in March with the solution, just since I was coming into the PR wanted to see if there where other feasible/no too hard options. But I understand your annoyance since you had the discussion multiple times. And because you where maintaining this and not me I was willing (still am) to ACK.

fjmolinas:
@kaspar030 the build filed, I can't see the details, might it be an autorization thing?

kaspar030:
> @kaspar030 the ~build~**test** filed, I can't see the details, might it be an autorization thing?

same here, I get 404. I've fixed some shellcheck complaints in hook/build and so triggered a rebuild, let's see.

kaspar030:
It passed now, but I still cannot see the dockercloud result.
(not that we configured it to do these builds in the first place)

fjmolinas (approved):
ACK, is surely possible to do it differently but this works now, lets improve later.

PR #69, opened 2019-03-19
add Dockerfile to build riotbuild image with opeoncd

fjmolinas:
### Contribution description

This PR adds a Dockerfile to add openocd to the riotbuild. This allow a user to compile, flash and access a terminal all threw docker. The only requirements would be adding udev rules to /etc/udev/rules.d.

It adds 180Mb to the docker image size, from 3.5Gb ---> 3.76Gb

For now you need to add "--privileged" when flashing, but this could be wrapped in the RIOT makefiles (RIOT-OS/RIOT#11220).

### Testing procedure

Build Image:

`docker build -t riotflash -f Dockerfile.openocd .
`

Compile and Flash:

`docker run --rm -i -t --privileged -v $(pwd):/data/riotbuild riotflash make -C /data/riotbuild/examples/hello-world BOARD=nucleo-l152re flash
`

Open terminal:

`docker run --rm -i -t --device=/tty/ACM0 --group-add plugdev  -v $(pwd):/data/riotbuild riotflash make -C /data/riotbuild/examples/hello-world BOARD=nucleo-l152re term
`

NOTE: the docker image in docker hub is 4 month old, not up to date with the latest changes changes is Dockerfile.
NOTE2: the commands above assume that the new image is riotflash, it could be riot/riotbuildv2 or whatever

### Issues/PRs references

#30

### DIFF
--- a/Dockerfile.openocd
+++ b/Dockerfile.openocd
@@ -1,0 +1,24 @@
+FROM riot/riotbuild
+
+# Install openocd dependencies
+RUN apt-get update && apt-get -y dist-upgrade \
+    && apt-get -y --no-install-recommends install \
+    libhidapi-hidraw0 \
+    libhidapi-dev \
+    libusb-1.0 \
+    pkg-config
+
+# Build openocd from source
+RUN mkdir -p opt \
+    && cd /opt \
+    && git clone --depth 1 https://github.com/ntfreak/openocd.git \
+    && cd openocd \
+    && ./bootstrap \
+    && ./configure --enable-stlink --enable-jlink --enable-ftdi --enable-cmsis-dap \
+    && make -j"$(nproc)" \
+    && make install-strip \
+    && cd .. \ 
+    && rm -rf openocd \
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /data/riotbuild
--- a/Dockerfile.openocd
+++ b/Dockerfile.openocd
@@ -6,6 +6,7 @@ RUN apt-get update && apt-get -y dist-upgrade \
     libhidapi-hidraw0 \
     libhidapi-dev \
     libusb-1.0 \
+    libudev-devi \
     pkg-config
 
 # Build openocd from source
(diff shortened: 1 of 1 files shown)


jcarrano:
I'm not sure about this. On the one hand, it is a _big_ plus to the user to not have to care about OpenOCD. On the other hand, many of the complications I have seen with new users are related to rules and configurations that they would normally not want to touch, so this part of the PR kind of negates the other benefits.

The main issue, though, is that you say the container must be run with more privileged, which normally is something to avoid. I do not quite understand why both `--device` and `--privileged` are needed. According to [these docs](https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities) `--privileged` encompasses all devices (and some extra privileges).

fjmolinas:
@jcarrano I was trying to figure out a work-around this and trying to integrate it into RIOT Makefiles [here](https://github.com/fjmolinas/RIOT/tree/docker_flash_term).  You actually don't need --device and --privileged at the same time. That is a mistake on my side, I have updated the description.

I still have two problems: 

1. --device works for using term using docker, not for flashing... still need privileged for this
2. when integrating into the makefile a user is created in the Dockerfile to create files in Host with same uid as host user. But this new user doesn't have the same acces rights and can't open the terminal (trying to figure out why),

Anyway the udev-rules problem is still present in the current riotbuild image.

fjmolinas:
@jcarrano I made a PR that could integrate this as an option in RIOT at https://github.com/RIOT-OS/RIOT/pull/11220, doesn't need privileged for using make term, but still does for make flash. What do you think?

Dockerfile.openocd
```diff
@@ -0,0 +1,25 @@
...
+    libhidapi-dev \
+    libusb-1.0 \
+    libudev-devi \
+    pkg-config
+
+# Build openocd from source
+RUN mkdir -p opt \
+    && cd /opt \
+    && git clone --depth 1 https://github.com/ntfreak/openocd.git \
+    && cd openocd \
```
kaspar030: commit should be pinned

PR #70, opened 2019-04-11
esp32: fix toolchain download

kaspar030:
The current way of first recursively cloneing the esp-idf-repo, then setting the wanted commit, doesn't work. Some of the submodules have advanced too far. The result is that building the current Dockerfile fails.

This PR fixes this by first cloning un-recursively, setting the wanted commit, then clone the submodules from a known state.


kaspar030:
&go

PR #62, opened 2018-11-24
riscv: Upgrade to newer toolchain release v7.2.0-4 20180606

gebart:
Removes the need to download a libfl.so.2 file from a different distro.

Only tested building, I don't own a RISC-V board so I can't test.

PR #71, opened 2019-04-11
update riscv toolchain to 8.2

kaspar030:
The current risc-v toolchain makes some applications not pass the tests.
This gcc 8.1 based one works fine. I didn't dig for reasons. My gut feeling is that the newly available newlib-nano saves enough stack space to not break the tests.

Supercedes #62.
~~Waiting for #70.~~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -134,21 +134,21 @@ ENV MIPS_ELF_ROOT /opt/mips-mti-elf/2016.05-03
 ENV PATH ${PATH}:${MIPS_ELF_ROOT}/bin
 
 # Install RISC-V binary toolchain
+ARG RISCV_VERSION=8.2.0-2.2-20190521
+ARG RISCV_BUILD=0004
 RUN mkdir -p /opt && \
-        wget -q https://github.com/gnu-mcu-eclipse/riscv-none-gcc/releases/download/v7.2.0-2-20180110/gnu-mcu-eclipse-riscv-none-gcc-7.2.0-2-20180111-2230-centos64.tgz -O- \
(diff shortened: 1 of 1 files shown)


aabadie:
There's already a [new release](https://github.com/gnu-mcu-eclipse/riscv-none-gcc/releases/tag/v8.2.0-2.2-20190521) of the toolchain. Maybe use this one ?

kaspar030:
- changed to 8.2
- parameterized toolchain version to easy future updates
- changed commit message to reflect new version

Dockerfile
```diff
@@ -130,21 +130,21 @@ ENV MIPS_ELF_ROOT /opt/mips-mti-elf/2016.05-03
...
     echo 'Deduplicating binaries' >&2 && \
     cd /opt/gnu-mcu-eclipse/riscv-none-gcc/*/riscv-none-embed/bin && \
-    for f in *; do rm "$f" && ln "../../bin/riscv-none-embed-$f" "$f"; done && cd -
+      for f in *; do test -f "../../bin/riscv-none-embed-$f" && \
+       ln -f "../../bin/riscv-none-embed-$f" "$f"; \
+      done && \
+    cd -
 
-# HACK download arch linux' flex dynamic library
-RUN wget -q https://sgp.mirror.pkgbuild.com/core/os/x86_64/flex-2.6.4-2-x86_64.pkg.tar.xz -O- \
```
fjmolinas: Why was this needed before and why is is not needed anymore?

kaspar030: See #39, libfl on ubuntu is somehow weird. the riscv toolchain could not use it. thus we used arch linux' libfl package.
The current risc-v toolchain comes bundled with libfl, thus the hack is not needed anymore.

kaspar030:
Thanks @fjmolinas! &go.

PR #73, opened 2019-07-08
add python dependencies for SUIT

kaspar030:
This PR adds needed dependencies for SUIT (https://github.com/RIOT-OS/RIOT/pull/11818)

To test, make sure docker is correctly installed.

Then:

- check out this PR
- build the container (eg. ```docker build -t riotdocker_dev .```)
- check out #11818
- from the checkout root, run ```riot (suit-pr)]$ docker run -ti --rm -v $(pwd):/data/riotbuild -u $(id -u) riotdocker_dev /bin/bash```
- in the container, ```cd examples/suit_update; make suit/publish```

It should succeed.
Otherwise this shouldn't affect anything, as this is only an addition of python packages.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,6 @@ flake8==3.5.0
 pexpect==4.2.1
 pycrypto==2.6.1
 pyserial==3.4
+ed25519==1.4
+cbor==1.0.0
+cryptography==2.6.1


kaspar030:
> Otherwise this shouldn't affect anything, as this is only an addition of python packages.

Basic functioning of the container is checked by Travis.

fjmolinas:
Re-based on #75 it now succeeds. So lets wait for #75.

emmanuelsearch:
@fjmolinas is this an ACK ;)
(now that #75 is merged)

kaspar030:
> Doing as suggested fixed the issue.

This has been resolved.

requirements.txt
```diff
@@ -4,3 +4,7 @@ flake8==3.5.0
 pexpect==4.2.1
 pycrypto==2.6.1
 pyserial==3.4
+ed25519==1.4
+cbor==1.0.0
+cryptography==2.6.1
+click==7.0
```
fjmolinas: This can be removed.

emmanuelsearch: @kaspar030 seems like an easy fix ;)
so let's merge this PR the last one preliminary to #11818 ?

kaspar030: done

fjmolinas:
GO!

PR #79, opened 2019-08-21
python: keep setuptools during installation of python packages

aabadie:
Should fix https://github.com/RIOT-OS/RIOT/issues/12058 when deployed on the workers. I'm also wondering if the `install_requirements.txt` file is really needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -65,6 +65,8 @@ RUN \
         python3 \
         python3-dev \
         python3-pip \
+        python3-setuptools \
+        python3-wheel \
         p7zip \
         rsync \
         ssh-client \
--- a/Dockerfile
+++ b/Dockerfile
@@ -96,13 +96,9 @@ RUN \
     apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # install required python packages from file
-COPY install_requirements.txt /tmp/install_requirements.txt
 COPY requirements.txt /tmp/requirements.txt
 RUN echo 'Installing python3 packages' >&2 \
-    && pip3 install --no-cache-dir -r /tmp/install_requirements.txt \
     && pip3 install --no-cache-dir -r /tmp/requirements.txt \
-    && pip3 uninstall -y -r /tmp/install_requirements.txt \
-    && rm /tmp/install_requirements.txt \
     && rm /tmp/requirements.txt
 
 # Install ARM GNU embedded toolchain
--- a/install_requirements.txt
+++ b/install_requirements.txt
@@ -1,2 +1,0 @@
-setuptools==39.0.1
-wheel==0.30.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,8 @@
 pyasn1==0.4.2
 ecdsa==0.13
 flake8==3.5.0
+setuptools==39.0.1
+wheel==0.30.0
 pexpect==4.2.1
 pycrypto==2.6.1
 pyserial==3.4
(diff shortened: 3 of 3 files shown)


jcarrano:
let's just remove install_requirements.txt altogether. It has only two packages (1 after this PR) and the way it is handled in the dockerfile (removing all install requirements) is pretty fragile.

If we leave the file there, then sooner or later someone will add more packages to it and we will have the same problems again.

aabadie:
Thanks! I was looking locally for a way to run the static-test using `BUILD_IN_DOCKER` but this is harder than I thought. The idea was to also run the static-test from travis here.

jcarrano:
> The idea was to also run the static-test from travis here.

that would be great. Another thing I was thinking is that installing the python package could perhaps moved to be the of lasts thing done in the Dockerfile. It seems to me that that changes more often than the other packages.

aabadie:
> Another thing I was thinking is that installing the python package could perhaps moved to be the of lasts thing done in the Dockerfile

see #80 

Regarding the possibility to run `static-test` using `BUILD_IN_DOCKER`, I think I need some advice from @cladmi here ;)

cladmi:
> Regarding the possibility to run `static-test` using `BUILD_IN_DOCKER`, I think I need some advice from @cladmi here ;)

Currently the command to do it is in `.travis.yml` and it could be mapped to a make target in the base `Makefile`

Something like this seems to work with

```
CI_BASE_BRANCH="riot/master" DOCKER="sudo docker" BUILD_IN_DOCKER=1 make static-test
```

As long as I am not within a worktree or anything.

```diff
diff --git a/.travis.yml b/.travis.yml
index bf08d3fa2..f8798c0b1 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,4 @@ before_script:
   - git fetch origin "${TRAVIS_BRANCH}:${TRAVIS_BRANCH}"

 script:
-  - docker run -a STDIN -a STDOUT -a STDERR --rm -u "$(id -u)"
-      -e CI_BASE_BRANCH="${TRAVIS_BRANCH}"
-      -v "${PWD}:/data/riotbuild" -v /etc/localtime:/etc/localtime:ro
-      riot/riotbuild make static-test
+  - CI_BASE_BRANCH=${TRAVIS_BRANCH} BUILD_IN_DOCKER=1 make static-test
diff --git a/makefiles/tests.inc.mk b/makefiles/tests.inc.mk
index fd5e492ce..b00465aea 100644
--- a/makefiles/tests.inc.mk
+++ b/makefiles/tests.inc.mk
@@ -1,3 +1,11 @@
-
+DOCKER ?= docker
+ifeq ($(BUILD_IN_DOCKER),1)
+static-test:
+       $(DOCKER) run -a STDIN -a STDOUT -a STDERR --rm -u "$(id -u)" \
+               -e CI_BASE_BRANCH="$(CI_BASE_BRANCH)" \
+               -v "$(PWD):/data/riotbuild" -v /etc/localtime:/etc/localtime:ro \
+               riot/riotbuild make static-test
+else # BUILD_IN_DOCKER
 static-test:
        ./dist/tools/static-tests.sh
+endif # BUILD_IN_DOCKER
```

If there is interest I could do the change correctly. It would require splitting `makefiles/docker.inc.mk` but if there are reviewers I could do it.

aabadie:
Thanks @cladmi. In fact, that was quite simple. I don't know why I haven't thought at looking at travis.yml file...

> If there is interest I could do the change correctly. It would require splitting makefiles/docker.inc.mk but if there are reviewers I could do it.

That would help I think. I'll review it when it's PRed.

PR #80, opened 2019-08-22
Dockerfile: move Python requirements installation at the end

aabadie:
This PR follows https://github.com/RIOT-OS/riotdocker/pull/79#issuecomment-523888269 from @jcarrano which makes good sense to me.
Since Python package requirements are evolving quite often, this will avoid docker to rebuild layer (like ESP32 toolchain) coming after Python packages installation in the Dockerfile.


aabadie:
Thanks for merging @miri64 !

PR #81, opened 2019-08-30
apt: remove not allowed dist-upgrade

Rotzbua:
* Base image should be up to date.

* "Avoid RUN apt-get upgrade and dist-upgrade, as many of the “essential” packages from the parent images cannot upgrade inside an unprivileged container. "

Reference:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run


Dockerfile
```diff
@@ -37,7 +37,7 @@ ENV LANG C.UTF-8
 RUN \
     dpkg --add-architecture i386 >&2 && \
     echo 'Upgrading all system packages to the latest available versions' >&2 && \
```
miri64: I guess this line must also be adapted then ;-)

kaspar030:
while this is "best practice", I'm not sure we actually should stop updating the base image. "ubuntu:bionic" is still at its initial version, right?

Rotzbua:
We should not update an underling layer, because it simply costs resources in build time and storage. The base image should be always up to date. 
The dockerfile uses the official docker image from ubuntu, which is regularly updated. Nevertheless docker use caches so docker may use an older local caches ubuntu base image. To avoid this behavior simply add `--pull` flag to ensure latest available version.

Rotzbua:
Should I add the `--pull` flag to travis script and: https://github.com/RIOT-OS/riotdocker/blob/2ac1e6c29b7c625de676b107c77a03e3cfffd41f/Dockerfile#L10
?

kaspar030:
> The dockerfile uses the official docker image from ubuntu, which is regularly updated.

Do you have a reference for this?

Docker Hub using old versions of the parent image might be the reason that every dist-upgrade actually upgrades *a lot*.

Rotzbua:
You mean?
https://docs.docker.com/docker-hub/official_images/

Ubuntu image:
https://hub.docker.com/_/ubuntu?tab=tags

> bionic25.48 MB
> Last updated5 days agobydoijanky

kaspar030:
Yes, thanks! I also found this: https://anchore.com/look-often-docker-images-updated/
That's seems fine.

kaspar030:
> Should I add the `--pull` flag to travis script and:

Yes please!

PR #82, opened 2019-09-11
add msp430-gcc-elf version 9.2

geromueller:
this commit adds the gcc 9.2 compiler for the msp430 targets.
inspired by https://aur.archlinux.org/packages/msp430-elf-gcc/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -187,6 +187,10 @@ ENV PATH $PATH:/opt/esp/xtensa-esp32-elf/bin
 # Install msp430-elf-gcc
 # see https://aur.archlinux.org/packages/msp430-elf-gcc
 
+# silence autotools/make
+ENV GNUMAKEFLAGS=--no-print-directory
+ENV V=0
+
 ENV MSP430_GCC_BINUTILS_VER=2.32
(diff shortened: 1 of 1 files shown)


kaspar030:
Thanks for working on this!

I'm thinking if we should split the building of toolchains in different repositories.
The compilation is quite slow, especially on the boxes that travis and docker hub provide for free.
I remember compiling cmake added *minutes* to travis times.
Also, once we have a recipe for building our own toolchains, we might want to actually do that for other architectures as well. In addition, we might want to make the resulting builds available as seperate downloads.

So I suggest creating a repo in github that has a .travis.yml building the msp430 toolchain and hook it up so on a merge to master, the resulting binaries get tar'ed up and made available as github download. Then in riotbuild, we just download the result.

I can assist with that!

kaspar030:
> So I suggest creating a repo in github that has a .travis.yml building the msp430 toolchain and hook it up so on a merge to master, the resulting binaries get tar'ed up and made available as github download. Then in riotbuild, we just download the result.

@geromueller I've finally done that. See #99. Could you give it a try?

kaspar030:
Closing in favor of #99.

kaspar030:
(Note to myself: compare used configure flags)

PR #88, opened 2019-09-23
codespell: install via pip instead of apt

aabadie:
Unfortunately the version of codespell installed by apt doesn't contain some of the options used in https://github.com/RIOT-OS/RIOT/pull/12232

This PR is installing codespell via pip3 with the latest current version (1.16) to be sure all required features can be used.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,7 +47,6 @@ RUN \
         ccache \
         cmake \
         coccinelle \
-        codespell \
         curl \
         cppcheck \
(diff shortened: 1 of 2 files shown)


aabadie:
The version installed is the right one: 
```
$ docker build . -t riot/riotbuild
$ docker run -ti --rm riot/riotbuild /bin/bash
root@3ea93c11e4c3:~# codespell --version
1.16.0
```

Issue #64, opened 2019-01-16
The container should have versioning and releases


jcarrano:
RIOT has versioning and regular releases. RIOT depends on this image as it is part of the build system.

Both RIOT _and_ the image are thoroughly tested on each release (of RIOT) so that we know RIOT builds with the particular version of riotdocker that was active when the release was done.

The problem is: _that image is not being saved anywhere_. In other words, the Docker integration in RIOT always uses latest, and _there is only latest because riotdocker does not have versions_.

What this means is:

- Past releases of RIOT may be broken as they use "latest" and latest keeps on being changed.
- The latest release of RIOT may well be broken the day after it is relased, just by updating the image (or even by just rebuilding it, since it does a `apt-get update && apt-get -y dist-upgrade`) And because the release is already out, it is not being tested.

What I propose:

- Have a version of riotdocker for each release of RIOT. These should not get changed except to fix bugs (though nobody is reporting bugs against a release.)
- Change RIOT to use the release-specific image.
- Have a master/latest image, than can build the master RIOT.
- Have a CI release, that can also build the master riot. Merge master into CI anb deploy once in a while.

### Steps:

- [x] Create branch
- [ ] Update build rules in Docker Hub.
- [ ] Come up with a testing/deployment procedure
- [ ] Update the docker integration in RIOT.

gebart:
I agree with your concerns, it is bad that we are only using the `latest` tag. Providing one tag per RIOT release sounds like a future-proof solution which should allow still building old releases for bug fix testing etc. A tag for the rolling development is also useful when working on development of the RIOT master branch.

jcarrano:
I created a new branch '2019.01-branch' at the point where we were at the last RIOT release (which is also the point at which we are now.)

What I need now is someone with access to our Docker Hub account configures the build so that the `riot/riotbuild:2019.01-branch` exists and builds from the new branch.

Next step is to update RIOT itself to use the image.

jcarrano:
Does anybody care? I insist: the way we are handling the docker image now means we can break the current release at any point, and previous releases are certainly broken. I do not think this is acceptable,

miri64:
Shall we start with this now, that 2019.04 is out?

miri64:
> Does anybody care? I insist: the way we are handling the docker image now means we can break the current release at any point, and previous releases are certainly broken. I do not think this is acceptable,

I care and agree, so please go ahead!

jcarrano:
[...]

PR #91, opened 2019-10-15
add TI msp430 gcc 8.3.0 ELF/newlib toolchain

kaspar030:
This PR adds the current official TI msp430 ELF toolchain.

@gschorcht @geromueller please take a look!

Compared to #82 , it avoids building the toolchain from source.
Compared to #62, this is using TI's current release, and does not install the support files.

PR to RIOT master updating msp430: RIOT-OS/RIOT#12457, which includes just the used msp430-support-files.

This PR will not do any compilation test on travis, as there's no code supporting this in master...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -198,6 +198,13 @@ RUN echo 'Installing ESP32 toolchain' >&2 && \
 
 ENV PATH $PATH:/opt/esp/xtensa-esp32-elf/bin
 
+ARG MSP430_URL=https://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/latest/exports
+ARG MSP430_VERSION=8.3.0.16_linux64
+RUN echo 'Installing TI MSP430 ELF toolchain' >&2 && \
+        wget -q ${MSP430_URL}/msp430-gcc-${MSP430_VERSION}.tar.bz2 -O- \
+            | tar -C /opt -xj
+ENV PATH $PATH:/opt/msp430-gcc-${MSP430_VERSION}/bin
+
 # install required python packages from file
 COPY requirements.txt /tmp/requirements.txt
 RUN echo 'Installing python3 packages' >&2 \


geromueller:
FYI: A couple of weeks ago I asked the riot-devs per email which version to use, the TI or vanilla one. The vote was leaning towards the vanilla version.

kaspar030:
> FYI: A couple of weeks ago I asked the riot-devs per email which version to use, the TI or vanilla one. The vote was leaning towards the vanilla version.

I asked the same, and the vote was more like, "why maintain our own toolchain compiles?". I was leaning towards vanilla, as TI's up to 8.2.0 needed patches to their own linker files.
But the new version works just fine.

Are there any concrete reasons for using the vanilla version?

gschorcht:
It seems that the installation of the Ubuntu packages could be removed, right?
https://github.com/RIOT-OS/riotdocker/blob/c6e22b82dec57bcc3b464c8d54d28c42127c6b45/Dockerfile#L90-L93

geromueller:
Maybe
- Use different compiler  flags for newlib
- Use and test any gcc version, even the latest

kaspar030:
> It seems that the installation of the Ubuntu packages could be removed, right?

Yeah, for now we'll need to keep this, as current master relies on it. For a while (until the release after  RIOT-OS/RIOT#12457), we'll have to keep both.

gschorcht:
I will take a look to PR https://github.com/RIOT-OS/RIOT/pull/12457 next, as soon as I have some time. At the biginning of the winter term, my schedule is always quite busy.

gschorcht:
I was looking for the votes you both mentioned [here](https://www.mail-archive.com/devel@riot-os.org/). 

@geromueller You had exactly two votes from other developers, but both of them were suggesting to use vanilla. Arguments were:
- More consistent to other toolchains (no TI specific stuff)
- Does not rely on TI for future updates (in case TI looses interest in maintaining their on version)

@kaspar030 I couldn't find your survey.

From my humble point of view, we shouldn't compile a toolchain by our self if the precompiled one from the vendor or any other source works and fulfills our requirements. It is not always necessary to have the latest gcc version. Compiling the toolchain takes a lot of time and resources to generate the Docker image. If I remember correctly, it may timeout during generation, but I'm not sure. And finally, if the vendor doesn't have interest anymore, then we can compile it by our self.

So, if version 8.3.0 works, fulfills our requirements and doesn't have the problems @kaspar030 mentioned for version 8.2.0, I would prefer to use the precompiled one.

I have tested it with https://github.com/RIOT-OS/RIOT/pull/12457 and it seems to work, at least the compilation. I will have a test whether produced code also works. The only thing I have found with the toolchain is that it obviously uses a newlib-nano (`tests/libc_newlib` fails) which requires to use module `newlib-nano` in the build system. But this can be fixed in https://github.com/RIOT-OS/RIOT/pull/12457.

gschorcht:
> using the TI one as default is perfectly fine for me. Nevertheless there may be occasions when the vanilla one is interesting too. We could use docker ARGS to be able to switch to the vanilla one if a developer want to compile it on his own?

Would gcc 9.2 require further adoptions in RIOT's source code for MSP430?

gschorcht:
> We could use docker ARGS to be able to switch to the vanilla one if a developer want to compile it on his own?

On the other hand, we might argue that the `riotdocker` image comes with the default toolchain version for developers who do not want to have effort to manually install a toolchain on their host and wanna to use the toolchain out of the box. Developers with very specific toolchain requirements would probably install the toolchain on their host anyway.

geromueller:
I do not understand the reluctance do add a feature some people are
interested in. If you have good reasons, remove it. If you don't, just
leave it in for those who want to test their code on different compilers
and platforms. That is what testing is all about, isn't it?

Gero

Am 18.10.19 um 11:02 schrieb Gunar Schorcht:

kaspar030:
> Would gcc 9.2 require further adaptations in RIOT's source code for MSP430?

AFAIK, no. I'm using a self-compiled msp430 (very similar to how @geromueller did it: started from the arch AUR build files, changed some newlib flags). It works fine.

kaspar030:
I'm actually all for compiling our own toolchains (even if just for verifying the vendor ones). But IMO that needs to be introduced properly.


1. We're relying on vendor provided toolchains for all the other architectures. TI used to have a bad toolchain story, but they have caught up (by [outsourcing](http://www.mittosystems.com/ti-msp430-gcc.html)) to a usable state. IMO, unless there's a reason ("we need compile flag X") and if TI's binaries work, we should use them

2. if we introduce this here in the docker container, there's basically no way for people to get the same toolchain compile. It'll be spread over the docker container. I could not download "the RIOT authorative msp430 toolchain version X" other than the newest one (by pulling the container and extracting it from there, or by building within the container).

3. doing the compilation within Docker makes the container build times go up *a lot*. Even cmake increased the build times (on Docker hub's public cloud and on travis) by *minutes*. A whole toolchain compile (bootstrap, first stage, second stage...) will be more. Everything twice, as the Murdock build container is built right after this one.

So I propose:

1. (assuming that TI's toolchain does work), go with the TI toolchain for now, make sure our msp430 port works with it
2. start a repository for toolchains (RIOT-OS/toolchains)
3. provide toolchain building scripts there (for all architectures, make them build as similar as possible)
4. make CI build those toolchains and make them available for download
5. gauge differences to the official ones
6. decide if we prefer to use our own binaries as the "official" ones

kaspar030:
> We could use docker ARGS to be able to switch to the vanilla one if a developer want to compile it on his own?

The whole point of the docker container is to bundle a somewhat stable toolchain.
If you are using a different one, `PATH=<path-to-your-toolchain/bin>:$PATH make ...` should be all you need, or am I missing something?

gschorcht:
> I'm actually all for compiling our own toolchains (even if just for verifying the vendor ones). But IMO that needs to be introduced properly.
> 
> 1. We're relying on vendor provided toolchains for all the other architectures. TI used to have a bad toolchain story, but they have caught up (by [outsourcing](http://www.mittosystems.com/ti-msp430-gcc.html)) to a usable state. IMO, unless there's a reason ("we need compile flag X") and if TI's binaries work, we should use them
> 2. if we introduce this here in the docker container, there's basically no way for people to get the same toolchain compile. It'll be spread over the docker container. I could not download "the RIOT authorative msp430 toolchain version X" other than the newest one (by pulling the container and extracting it from there, or by building within the container).
> 3. doing the compilation within Docker makes the container build times go up _a lot_. Even cmake increased the build times (on Docker hub's public cloud and on travis) by _minutes_. A whole toolchain compile (bootstrap, first stage, second stage...) will be more. Everything twice, as the Murdock build container is built right after this one.
> 
> So I propose:
> 
> 1. (assuming that TI's toolchain does work), go with the TI toolchain for now, make sure our msp430 port works with it
> 2. start a repository for toolchains (RIOT-OS/toolchains)
> 3. provide toolchain building scripts there (for all architectures, make them build as similar as possible)
> 4. make CI build those toolchains and make them available for download
> 5. gauge differences to the official ones
> 6. decide if we prefer to use our own binaries as the "official" ones



> So I propose:
> 
> 1. (assuming that TI's toolchain does work), go with the TI toolchain for now, make sure our msp430 port works with it
> 2. start a repository for toolchains (RIOT-OS/toolchains)
> 3. provide toolchain building scripts there (for all architectures, make them build as similar as possible)
> 4. make CI build those toolchains and make them available for download
> 5. gauge differences to the official ones
> 6. decide if we prefer to use our own binaries as the "official" ones

I completely agree with @kaspar030.

gschorcht:
From my point of view, we should provide `riotdocker` for only one toolchain version for which we know that it is working with RIOT-OS. We should always use precompiled archives, our own or the official ones. The dockerfile in #82 seems to be much more complex than what we have for all other platforms and it will take a lot of time to generate the image.

gschorcht:
To make some progress with the new toolchain and https://github.com/RIOT-OS/RIOT/pull/12457, I will merge this first. Regarding the PR #82, we will find a solution according to @kaspar030's proposal in https://github.com/RIOT-OS/riotdocker/pull/91#issuecomment-543637485.

gschorcht:
@kaspar030 Would you like to rebase it before the merge?

gschorcht:
@kaspar30 Could you rebase your branch to the current master and force it? I have accidentally let GitHub merge the master branch into your branch :dark_sunglasses:

kaspar030:
> Could you rebase your branch to the current master and force it?

yup, done.

gschorcht:
@kaspar030 Thanks for contributing it. Once it is rolled out to Murdock, we can continue with PR https://github.com/RIOT-OS/RIOT/pull/12457.

gschorcht:
> So I propose:
> 
> 1. (assuming that TI's toolchain does work), go with the TI toolchain for now, make sure our msp430 port works with it
> 2. start a repository for toolchains (RIOT-OS/toolchains)
> 3. provide toolchain building scripts there (for all architectures, make them build as similar as possible)
> 4. make CI build those toolchains and make them available for download
> 5. gauge differences to the official ones
> 6. decide if we prefer to use our own binaries as the "official" ones
 
This might be related to issue #64 which proposes the versioning of riotdocker containers for releases. This would become easy with a toolchain repository.

PR #93, opened 2019-10-19
esp8266: toolchain update for the new RTOS SDK

gschorcht:
This PR adds a new ESP8266 toolchain that uses a newer GCC version and the new ESP8266 RTOS SDK instead of the old ESP8266 NONOS SDK. It is required for the reimplementation of the ESP8266 RIOT port in PR https://github.com/RIOT-OS/RIOT/pull/11108.

This new toolchain is installed in parallel to the old one to be able to compile the current master but also to compile the PR  https://github.com/RIOT-OS/RIOT/pull/11108. Once PR https://github.com/RIOT-OS/RIOT/pull/11108 is merged, the old toolchain can be removed.

The additional environment variable settings for the old toolchain are required to be able to divide the toolchain in RIOT's build system.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -172,6 +172,34 @@ RUN echo 'Installing ESP8266 toolchain' >&2 && \
     rm -rf .git
 
 ENV PATH $PATH:/opt/esp/esp-open-sdk/xtensa-lx106-elf/bin
+ENV ESP8266_SDK_DIR /opt/esp/esp-open-sdk/sdk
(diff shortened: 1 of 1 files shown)


gschorcht:
> ACK.

Thanks :smile:
 
> Tested basic compilation for esp8266-esp-12x, which confirms the toolchain basically works.
> This is not used by master (yet), so any problems will show up when merging RIOT/pull/11108.

Don't worry, be sure that I have tested this parallel installation of old and new toolchain with master and https://github.com/RIOT-OS/RIOT/pull/11108.

PR #95, opened 2019-12-22
esp8266: replacement of SDK library printf functions

gschorcht:
### Description

This PR just references another commit [c0174ef](https://github.com/gschorcht/RIOT-Xtensa-ESP8266-RTOS-SDK/commit/c0174eff7278eb5beea66ce1f65b7af57432d2a9) of the ESP8266 toolchain repository which simply contains a number of function call replacements in SDK libraries.

### Background

In order to be able to control the log output from ESP8266 SDK libraries according to the defined log level, a number of library-specific `printf` functions were implemented in RIOT with PR [#12998](https://github.com/RIOT-OS/RIOT/pull/12998) for use by the SDK. These library-specific `printf` functions intercept the log output of the SDK and redirect it to the RIOT log macros.

To get these SDK library-specific functions working, all `printf`/`ets_printf` calls in SDK libraries have to be replaced by the calls of the library-specific `printf` functions.

### Dependencies

Depends on PR [#12998](https://github.com/RIOT-OS/RIOT/pull/12998) (PR [#12998](https://github.com/RIOT-OS/RIOT/pull/12998) has to be merged before this change can be merged. Otherwise, compilation errors of RIOT for ESP8266 will occur.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -190,13 +190,14 @@ RUN echo 'Installing ESP8266 toolchain' >&2 && \
     cd /opt/esp && \
     git clone https://github.com/gschorcht/RIOT-Xtensa-ESP8266-RTOS-SDK.git ESP8266_RTOS_SDK && \
     cd ESP8266_RTOS_SDK/ && \
-    git checkout -q f074414c0705715a44b8e59d53b03d90b7630382 && \
-    rm -rf .git* docs examples make tools && \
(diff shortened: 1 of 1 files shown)


gschorcht:
@kaspar030 Since PR https://github.com/RIOT-OS/RIOT/pull/12998 was released with 2020.01, we should change now the toolchain to make it effective for ESP8266.

gschorcht:
Thanks for merging.

PR #97, opened 2020-01-16
esp32: update for toolchain version v2

gschorcht:
### Description

This PR changes the reference to another commit of the [precompiled ESP32 toolchain repository](https://github.com/gschorcht/xtensa-esp32-elf) to update the toolchain to a version that is required to fix issue https://github.com/RIOT-OS/RIOT/issues/13133.

### Test

Compilation of all examples/tests in current RIOT master should succeed.

### Dependencies

Requires PR https://github.com/RIOT-OS/RIOT/pull/13141

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -222,7 +222,7 @@ RUN echo 'Installing ESP32 toolchain' >&2 && \
     cd /opt/esp && \
     git clone https://github.com/gschorcht/xtensa-esp32-elf.git && \
     cd xtensa-esp32-elf && \
-    git checkout -q ca40fb4c219accf8e7c8eab68f58a7fc14cadbab
+    git checkout -q 9fa454fc178136bc2828a667c73b2667383545e5
 
 ENV PATH $PATH:/opt/esp/xtensa-esp32-elf/bin
 


kaspar030:
rollout will have to wait until after release testing, though...

gschorcht:
Wow, very fast. Thanks.

So we have also to merge and backport https://github.com/RIOT-OS/RIOT/pull/13141. Otherwise, a user who is generating docker image from the Dockerfile would run into compilation errors.

> rollout will have to wait until after release testing, though...

The current RIOT master should work with the updated toolchain too.

kaspar030:
> The current RIOT master should work with the updated toolchain too.

Yeah, it's just that during these weeks, I try to change as little as possible in the CI infrastructure. Losing half a day on a revert-and-redeploy-and-retest cycle is painful for the people trying to get backports and fixes CIed. (Not that I think this PR would cause that.)

gschorcht:
@kaspar030 The release is out. We could try to roll out that change now.

The fix in PR https://github.com/RIOT-OS/RIOT/pull/13141 was backported to the release with PR https://github.com/RIOT-OS/RIOT/pull/13144. So there should be no risk to roll it out. Once it is rolled out, we can revert https://github.com/RIOT-OS/RIOT/pull/13140.

PR #98, opened 2020-01-29
Dockerfile: install python-protobuf

benpicco:
Required for nanopb-0.4.0 (https://github.com/RIOT-OS/RIOT/pull/13228)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,6 +73,7 @@ RUN \
         pcregrep \
         protobuf-compiler \
         python \
+        python-protobuf \
         python3 \
         python3-dev \
         python3-pip \


kaspar030:
This would install the python2 version, which is probably outdated.
The python3 version is installed here:

https://github.com/RIOT-OS/riotdocker/blob/13929cddd17cfefee21c8160f08d7e9a95ddc8b9/requirements.txt#L13

Maybe we should just update that?

benpicco:
The thing is I can't get nanopb-0.4.0 working with the python3 version.

benpicco:
```
         ********************************************************************
         *** Failed to import the protocol definitions for generator.     ***
         *** You have to run 'make' in the nanopb/generator/proto folder. ***
         ********************************************************************
    
Traceback (most recent call last):
  File "/home/benpicco/dev/RIOT-staging/tests/pkg_nanopb/bin/pkg/blackpill/nanopb/generator/nanopb_generator.py", line 42, in <module>
    from .proto import nanopb_pb2
ModuleNotFoundError: No module named '__main__.proto'; '__main__' is not a package
```

Is what I get. The error disappears when I use `python` instead of `python3`.
Now I don't know enough about Python to properly debug this issue.

benpicco:
not needed anymore.

miri64:
An explanation why it is not needed anymore would be nice in case somebody else falls into the same rabbit hole as you did. ;-)

benpicco:
In the meantime nanopb-0.4.1 was tagged which includes [this](https://github.com/nanopb/nanopb/commit/96670dad4df24f1a1f1f65b302dc64f1216650ed) fix.

PR #99, opened 2020-02-03
msp430-elf: switch to RIOT-OS/toolchains 9.2.0-9

kaspar030:
This PR switches the (currently not used because it is broken) TI msp430-elf toolchain with the newly created one from https://github.com/RIOT-OS/toolchains/releases.

This currently won't be tested by travis as there's no corresponding code in master.
I tested building with https://github.com/RIOT-OS/RIOT/pull/12457, works fine.

Updates / fixes #91.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -226,12 +226,17 @@ RUN echo 'Installing ESP32 toolchain' >&2 && \
 
 ENV PATH $PATH:/opt/esp/xtensa-esp32-elf/bin
 
-ARG MSP430_URL=https://software-dl.ti.com/msp430/msp430_public_sw/mcu/msp430/MSPGCC/latest/exports
-ARG MSP430_VERSION=8.3.0.16_linux64
-RUN echo 'Installing TI MSP430 ELF toolchain' >&2 && \
-        wget -q ${MSP430_URL}/msp430-gcc-${MSP430_VERSION}.tar.bz2 -O- \
-            | tar -C /opt -xj
-ENV PATH $PATH:/opt/msp430-gcc-${MSP430_VERSION}/bin
+# RIOT toolchains
+ARG RIOT_TOOLCHAIN_GCC_VERSION=9.2.0
+ARG RIOT_TOOLCHAIN_PACKAGE_VERSION=9
+ARG RIOT_TOOLCHAIN_TAG=20200203165626-106c6b8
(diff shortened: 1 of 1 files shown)


kaspar030:
- fixed travis, rebased

gschorcht:
Makes PR #82 obsolete, right?

kaspar030:
> Makes PR #82 obsolete, right?

Yeah. Building the toolchain alone takes ~25-30min on Travis, so having the build as part of this container build just doesn't work.

Dockerfile
```diff
@@ -226,12 +226,17 @@ RUN echo 'Installing ESP32 toolchain' >&2 && \
...
+ARG RIOT_TOOLCHAIN_GCC_VERSION=9.2.0
+ARG RIOT_TOOLCHAIN_PACKAGE_VERSION=9
+ARG RIOT_TOOLCHAIN_TAG=20200203165626-106c6b8
+ARG RIOT_TOOLCHAIN_GCCPKGVER=${RIOT_TOOLCHAIN_GCC_VERSION}-${RIOT_TOOLCHAIN_PACKAGE_VERSION}
+ARG RIOT_TOOLCHAIN_SUBDIR=${RIOT_TOOLCHAIN_GCCPKGVER}-${RIOT_TOOLCHAIN_TAG}
+
+ARG MSP430_URL=https://github.com/RIOT-OS/toolchains/releases/download/${RIOT_TOOLCHAIN_SUBDIR}/riot-msp430-elf-${RIOT_TOOLCHAIN_GCCPKGVER}.tgz
+RUN echo 'Installing RIOT MSP430 ELF toolchain' >&2 && \
+        wget -q ${MSP430_URL} -O- | tar -C /opt -xz
+ENV PATH $PATH:/opt/riot-toolchain/msp430-elf/${RIOT_TOOLCHAIN_GCCPKGVER}/bin
```
gschorcht: @kaspar030 There is a problem with the `PATH` variable and the directory extracted from the package. While the sub-directory `/opt/riot-toolchain/msp430-elf/9.2.0` doesn't contain the package version in its name, the `PATH` variable is set to `PATH=...:/opt/riot-toolchain/msp430-elf/9.2.0-9`. Therefore, `msp430-elf-*` tools aren't found without setting the PATH variable.

gschorcht: Either we use the `PATH` with out the package version, or the package generation has to be modified.

kaspar030: >  the package generation has to be modified.

I'll go with that. The package version should be in there, otherwise updates might "mix" versions when blindly extracting new versions.

PR #103, opened 2020-04-17
Dockerfile: install afl-gcc

benpicco:
`afl-gcc` is needed for automatic build of fuzz tests.

https://github.com/RIOT-OS/RIOT/pull/13157#issuecomment-615264105


benpicco:
Not needed after all

PR #104, opened 2020-05-05
Update to Ubuntu 2020.04 LTS

kaspar030:
This PR updates the base for our build container to Ubuntu focal (2020.04).

There's no coccinelle package anymore for focal in the coccinelle ppa, so I've added a binary package.
Same with libsocketcan.

In order to not clutter the git history with binary data, those package files have been added using git lfs, and the build.yml workflow has been adapted to pull the lfs files.

~~WIP, want to see travis output~~

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.deb filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -42,6 +42,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+        with:
+          lfs: 'true'
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -26,6 +26,14 @@ LABEL maintainer="Kaspar Schleiser <kaspar@riot-os.org>"
 
 ENV DEBIAN_FRONTEND noninteractive
 
+ENV LC_ALL C.UTF-8
+ENV LANG C.UTF-8
+
+# copy some included packages
+RUN mkdir /pkgs
+COPY files/libsocketcan-dev_0.0.11-1_i386.deb /pkgs/libsocketcan-dev_0.0.11-1_i386.deb
+COPY files/libsocketcan2_0.0.11-1_i386.deb /pkgs/libsocketcan2_0.0.11-1_i386.deb
+
 # The following package groups will be installed:
 # - update the package index files to latest available version
 # - native platform development and build system functionality (about 400 MB installed)
@@ -103,12 +111,10 @@ RUN \
         # Filter by symlinks starting with /usr/bin/llvm-${LLVM_VERSION}
         case "${SYMTARGET}" in "/usr/lib/llvm-${LLVM_VERSION}"* ) ln -sf ${SYMTARGET} ${SYMNAME}; esac \
     done \
-    && echo 'Installing socketCAN' >&2 && \
-    apt-get -y --no-install-recommends install \
-        libsocketcan-dev:i386 \
-        libsocketcan2:i386 \
+    && echo 'Installing local packages' >&2 && \
+        apt install -y --no-install-recommends /pkgs/*.deb \
     && echo 'Cleaning up installation files' >&2 && \
-    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /pkgs
 
 # Install ARM GNU embedded toolchain
 # For updates, see https://developer.arm.com/open-source/gnu-toolchain/gnu-rm/downloads
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -94,9 +94,9 @@ RUN \
         avr-libc \
     && echo 'Installing LLVM/Clang toolchain' >&2 && \
     apt-get -y --no-install-recommends install \
-        llvm-${LLVM_VERSION}/bionic-updates \
-        clang-${LLVM_VERSION}/bionic-updates \
-        clang-tools-${LLVM_VERSION}/bionic-updates \
+        llvm-${LLVM_VERSION} \
+        clang-${LLVM_VERSION} \
+        clang-tools-${LLVM_VERSION} \
         llvm \
         clang \
         clang-tools \
--- a/riotbuild/files/libsocketcan-dev_0.0.11-1_i386.deb
+++ b/riotbuild/files/libsocketcan-dev_0.0.11-1_i386.deb
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:4e5d32040edb35388f4492ea4db10a00010e942a2139a20ede0b1e4d2e027e2f
+size 7588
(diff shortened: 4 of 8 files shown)


kaspar030:
2020.04 doesn't include coccinelle. This PR adds a self-compiled version, need to test that. And, maybe not include it in the repo, it is like 5mb.

What do you guys think?

aabadie:
Ubuntu 20.04 also includes a version of doxygen that generates tons of warning. This will break the static tests.

kaspar030:
> Ubuntu 20.04 also includes a version of doxygen that generates tons of warning. This will break the static tests.

oh yes, I got the same on arch. That's a bummer.

miri64:
This PR should also be tested for https://github.com/RIOT-OS/RIOT/issues/14288.

miri64:
> > Ubuntu 20.04 also includes a version of doxygen that generates tons of warning. This will break the static tests.
> 
> oh yes, I got the same on arch. That's a bummer.

I guess its time to find out how to deactivate those in config ;-).

MrKevinWeiss:
Do you need to add `apt install linux-libc-dev:i386` as well so native can compile?

kaspar030:
> Do you need to add `apt install linux-libc-dev:i386` as well so native can compile?

not sure it get's selected somehow, but travis did compile native correctly.

kaspar030:
- rebased. let's see if the static tests fail now (they should show a lot of new documentation warnings)

kaspar030:
>     * they should show a lot of new documentation warnings

Hm, it didn't. :/

miri64:
> Ubuntu 20.04 also includes a version of doxygen that generates tons of warning. This will break the static tests.

As those checks are run through a parsing script anyways, we could add exclude patterns for the warnings we deem false negatives.

miri64:
Many of those false positives should have been fixed in [doxygen 1.8.16](https://github.com/doxygen/doxygen/commit/7a10b6e784bf1ed882157471038e174df850b06c) btw... not sure, why they still show up. Will investigate.

Dockerfile
```diff
@@ -29,6 +29,13 @@ ENV DEBIAN_FRONTEND noninteractive
 ENV LC_ALL C.UTF-8
 ENV LANG C.UTF-8
 
+# copy some included packages
+RUN mkdir /pkgs
+COPY files/coccinelle_1.0.8~19.04npalix1_amd64.deb /pkgs/coccinelle_1.0.8~19.04npalix1_amd64.deb
```
benpicco, line 34: Why not use https://packages.ubuntu.com/bionic/coccinelle ?

Dockerfile
```diff
@@ -29,6 +29,13 @@ ENV DEBIAN_FRONTEND noninteractive
...
+COPY files/libsocketcan-dev_0.0.11-1_i386.deb /pkgs/libsocketcan-dev_0.0.11-1_i386.deb
```
benpicco, line 35: and https://packages.ubuntu.com/bionic/libsocketcan-dev

benpicco: ah nvm, this particular version only appeared in groovy

Dockerfile
benpicco: ah nvm, this particular version only appeared in groovy

larseggert:
Any ETA on merging this? I'm trying to port a package to RIOT that requires a pretty modern `cmake`.

kaspar030:
This is mostly blocked by the newer Doxygen, which warns on a lot more undocumented stuff than the bionic one.
Fixing all of those is a bigger task. We could work around by using the older version for this container.
Anyhow, there's no ETA and AFAIK, noone working on this.

miri64:
> This is mostly blocked by the newer Doxygen, which warns on a lot more undocumented stuff than the bionic one.

Is there really no one willing to look into how to surppress or fix those doxygen warnings? Looking at the doxygen output in Arch I don't see this solving by itself. We are now closer to the next LTS than we are to 2020.04, so only ~1,5 years time to solve this before 2018.04 will reach EOL. IMHO someone(TM) should fix this ASAP. It's a boring job and shooting at a moving target, granted, but it should be done.

miri64:
RIOT-OS/RIOT#16779 was merged so there should be nothing in the way for merging this now. @kaspar030 could you rebase / adopt for current master, please?

chrysn:
Now that @miri64 ran the script of https://github.com/RIOT-OS/RIOT/pull/16779, there's some urgency to this -- while that is in but this is not, there's a game of whack-a-mole to be played with any PR being merged that introduces documentation changes that don't trigger in CI but in the new CI would be fatal. (For example, judging from the output of my local Doxygen, the changes of https://github.com/RIOT-OS/RIOT/commit/d74bfaac17a0ae1069b5dbc5ff03fa49326d73b4 might be problematic).

kaspar030:
on it

kaspar030:
Seems like there are doxygen warnings not whitelisted in current master:
```
ERROR: Doxygen generates the following warnings:
	sys/include/net/cord/config.h:115: warning: unbalanced grouping commands
	boards/adafruit-clue/include/board.h:108: warning: Member ILI9341_PARAM_ROTATION (macro definition) of file board.h is not documented.
	boards/esp32-wrover-kit/include/board.h:122: warning: Member ILI9341_PARAM_ROTATION (macro definition) of file board.h is not documented.
	boards/pinetime/include/board.h:79: warning: Member ILI9341_PARAM_ROTATION (macro definition) of file board.h is not documented.
	cpu/arm7_common/include/irq_arch.h:73: warning: Member irq_is_enabled(void) (function) of file irq_arch.h is not documented.
	cpu/msp430_common/include/irq_arch.h:91: warning: Member irq_is_enabled(void) (function) of file irq_arch.h is not documented.
	cpu/riscv_common/include/irq_arch.h:100: warning: Member irq_is_enabled(void) (function) of file irq_arch.h is not documented.
	drivers/ili9341/include/ili9341_params.h:63: warning: Member ILI9341_PARAM_ROTATION (macro definition) of file ili9341_params.h is not documented.

kaspar030:
> (For example, judging from the output of my local Doxygen, the changes of [RIOT-OS/RIOT@d74bfaa](https://github.com/RIOT-OS/RIOT/commit/d74bfaac17a0ae1069b5dbc5ff03fa49326d73b4) might be problematic).

yup ;)

kaspar030:
>  there's some urgency to this

I think this will need some coordination, and we'll have to use current master (vs. last stable RIOT) for the riotdocker tests. Also, merging this will break static tests for the current release branch. I propose we tackle this at tomorrow's sprint day.

miri64:
> I think this will need some coordination, and we'll have to use current master (vs. last stable RIOT) for the riotdocker tests. Also, merging this will break static tests for the current release branch. I propose we tackle this at tomorrow's sprint day.

Shouldn't version-based tagging now be possible? At least for the 2020.07-branch there is an associated branch here. Shouldn't be too difficult to find the docker version with which 2020.10, 2021.01, 2021.04 and 2021.07 were tested and built.

kaspar030:
bors try

kaspar030:
doccheck succeeded now.

> In some of my experiments on my 20.04 machine I noticed some problems with tinydtls that I do not have on my Arch machine (some numbers are wrong on the receiving end). Maybe we should check that as well before merging.

do you remember any details?

kaspar030:
> do you remember any details?

we're early in the release cycle, so I'd suggest now that the doxygen issue is solved, we go with this as is. for tinydtls, we can open an issue so we don't forget verifying that still works. (it should, the toolchain used is the same).

fjmolinas:
Please squash

miri64:
I lately noticed some weird behavior of gCoAP over TinyDTLS when compiled my Ubuntu 20.04 machine (the python-based server receives TinyDTLS packets from the gCoAP client that it can not decypher, due to an "unknown TLS extension 23"). Not a blocker, but we should keep an eye of this during the coming release cycle.

miri64:
bors try

fjmolinas (approved):
ACK

fjmolinas:
bors merge

larseggert:
Thank you for getting this merged!

PR #108, opened 2020-07-07
requirements.txt: add riotctrl as a dependency

miri64:
This adds [`riotctrl`](https://github.com/RIOT-OS/riotctrl/) to the docker image.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,3 +16,4 @@ scikit-learn==0.22.1
 joblib==0.14.0
 emlearn==0.10.1
 jinja2==2.11.2
+riotctrl>=0.1.1


requirements.txt
```diff
@@ -14,3 +14,4 @@ protobuf==3.10.0
 scikit-learn==0.22.1
 joblib==0.14.0
 emlearn==0.10.1
+git+https://github.com/RIOT-OS/riotctrl
```
aabadie, line 17: Would it makes sense if we first release it on pypi and use this version instead ?

miri64: Does it make sense to release it on pypi if most likely we will only use this internally?

aabadie: For compatibility between versions, I think so. If the ifconfig output of a given version of RIOT is changed, then previous RIOT version will always fail. Having a specific RIOT version linked to a riotctrl version would prevent that. And we would also have to use this version the Dockerfile and link each RIOT version to a Docker image version. This we are not doing for the moment.

miri64: > For compatibility between versions, I think so. If the ifconfig output of a given version of RIOT is changed, then previous RIOT version will always fail. Having a specific RIOT version linked to a riotctrl version would prevent that. And we would also have to use this version the Dockerfile and link each RIOT version to a Docker image version. This we are not doing for the moment.

`riotctrl_shell` is bound to the RIOT version, not the `riotctrl` version. Thus `ifconfig`'s output is bound to RIOT, not `riotctrl`. So I'm not sure what you mean.

miri64: We could also link a tag here, if you want to link to a static version.

aabadie: > So I'm not sure what you mean

You are right, my assumption was wrong. I think that I had in mind to move all shell interation code from RIOT to riotctrl. But maybe that's useless.

miri64: No, that was never the idea and that is not how I remember the discussions about https://github.com/RIOT-OS/RIOT/pull/10949 concluded.

aabadie: > We could also link a tag here, if you want to link to a static version.

We could publish on pypi for each new tag of riotctrl :)

miri64: (for the very same reason we draw out here: it is better to keep the input/output handling for specific shell commands with the shell commands)

miri64: > > We could also link a tag here, if you want to link to a static version.
> 
> We could publish on pypi for each new tag of riotctrl :)

We can, but again: I don't see the point of that for a very RIOT-specific library.

miri64: After some reconsideration [`riotctrl`](https://pypi.org/project/riotctrl/) is now on PyPI (adding it as a dependency in `setup.py` is harder with github links I believe). I changed the line accordingly.

kaspar030:
pls rebase and squash

benpicco:
Please squash (& rebase)

miri64:
Rebased (rebased!) and squashed (_aaand_ squashed)

(and updated `riotctrl` minimal version)

PR #109, opened 2020-07-07
picolibc: Add initial support for picolibc

bergzand:
Add support for picolibc in the docker image.

This builds the picolibc implementation for arm, esp32 and riscv32. The riscv implementation seems to work with RIOT-os/RIOT#12305, but I haven't tested it yet on actual hardware as I don't own one of the hifive boards.

Meson is grabbed from pip and not from the bionic repos because the version in bionic is too old (0.45 vs 0.50)

An alternative approach would be to simply grab the debian buster deb's and install those. Please let me know what is preferred here.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,6 +70,7 @@ RUN \
         libpcre3 \
         libtool \
         m4 \
+        ninja-build \
         parallel \
         pcregrep \
         protobuf-compiler \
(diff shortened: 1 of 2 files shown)


aabadie:
I tested the build in Docker using this PR and the PR in RIOT. It works on samr21 and simplify things when one just want to do a quick test.

aabadie:
Do you think the `cross-riscv-none-embed.txt` file could be contributed upstream (with a do-riscv-none-embed-configure script similar to other architecture) ?

kaspar030:
please rebase

bergzand:
> please rebase

Rebased

kaspar030:
> Rebased

and please squash. the way I see it, this is not used in master, so we cannot break much.

bergzand:
Squashed!

benpicco:
Would be great if we can test the picolibc code paths in CI :)

Not sure how long this takes to compile, but there is also a [Debian package](https://packages.debian.org/search?keywords=picolibc&searchon=names&suite=testing&section=all).

kaspar030 (approved):
ACK.

kaspar030:
Let's see...

benpicco:
Looks like there is a conflict with the RISC-V toolchain: [examples/gnrc_minimal/hifive1:gnu.txt](https://ci.riot-os.org/RIOT-OS/RIOT/15011/ec6d8d04cc2875c199c8ef5f71c99a793c8f1dee/output/compile/examples/gnrc_minimal/hifive1:gnu.txt)

benpicco:
When I try to build https://github.com/RIOT-OS/RIOT/pull/15011 with `BUILD_IN_DOCKER=1` I get

```
/data/riotbuild/riotbase/sys/picolibc_syscalls_default/syscalls.c:236:5: error: implicit declaration of function 'FDEV_SETUP_STREAM' [-Werror=implicit-function-declaration]
  236 |     FDEV_SETUP_STREAM(picolibc_put, picolibc_get, picolibc_flush, _FDEV_SETUP_RW);
      |     ^~~~~~~~~~~~~~~~~
/data/riotbuild/riotbase/sys/picolibc_syscalls_default/syscalls.c:236:67: error: '_FDEV_SETUP_RW' undeclared here (not in a function)
  236 |     FDEV_SETUP_STREAM(picolibc_put, picolibc_get, picolibc_flush, _FDEV_SETUP_RW);
      |                                                                   ^~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

This does not happen when I use the host-installed picolibc from the Debian package.

PR #113, opened 2020-07-23
requirements.txt: add dependency to jinja2 template engine

aabadie:
It will be required by https://github.com/RIOT-OS/RIOT/pull/14595 and it's not available in the Docker image.


aabadie:
rebased

PR #114, opened 2020-09-30
requirements.txt: add pyhsslms

benpicco:
The upstream suit-manifest-generator [requires](https://github.com/ARMmbed/suit-manifest-generator/blob/master/setup.py#L55) the pyhsslms package.
In order to use the upstream version as a package, also include this in riotdocker.

Needed for https://github.com/RIOT-OS/RIOT/pull/15095


kaspar030 (approved):
ACK.

PR #115, opened 2020-09-30
llvm: Bump to LLVM version 10

bergzand:
Bump our LLVM/Clang to version 10. Explicitly adding the version number so that we don't accidentally start to pull in v11 when bionic-updates bumps the available version.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,7 @@ ENV LANG C.UTF-8
 # allow the cleanup to actually save space.
 # Total size without cleaning is approximately 1.525 GB (2016-03-08)
 # After adding the cleanup commands the size is approximately 1.497 GB
+ARG LLVM_VERSION=10
 RUN \
     dpkg --add-architecture i386 >&2 && \
     echo 'Update the package index files to latest available versions' >&2 && \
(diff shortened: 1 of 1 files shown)


kaspar030:
travis failed on the llvm tests.

bergzand:
`/usr/bin/clang` is not provided by the `clang-10` package.

benpicco:
Do we need to run `update-alternatives` or can we just create the symlink ourself?

bergzand:
> Do we need to run `update-alternatives` or can we just create the symlink ourself?

I think we have to create it manually. `/usr/bin/{llvm,clang}` are part of the `llvm` and `clang` package I think.

bergzand:
I've added a bit of logic to add symlinks based on already existing symlinks in /usr/bin pointing to /usr/lib/llvm-10 and recreate them without the version suffix. This should override all the llvm/clang symlinks to use version 10.

bergzand:
This is ready for review:

```
koen@morgen ~/dev/riotdocker $ docker run --rm -i -t -u $UID  riotbuild:custom /usr/bin/clang --version
clang version 10.0.0-4ubuntu1~18.04.2 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

bergzand:
A different approach is to download the binary releases from the LLVM project, verify them and install them manually in `/usr/local/bin`. That allows a bit more freedom in that we're not restricted to what Ubuntu provides in their updates repo.

kaspar030:
Lgtm. Is there no update-alternatives for llvm?

bergzand:
> Is there no update-alternatives for llvm?

I would prefer to use that over the manual creation of the symlinks. Please prove me wrong, but I could not find it. I did find a few ways to manually add entries to the update-alternatives by specifying the link name, target and the priority, but that doesn't really help here.

bergzand:
Can I squash this?

kaspar030:
please squash!

kaspar030 (approved):
ACK.
Basic compilation is asserted by travis, change LGTM.

PR #116, opened 2020-09-30
Dockerfile: install picolibc & ARM/RISC-V GCC via package manager

benpicco:
There seems to be a problem with the picolibc install in riotdocker. (https://github.com/RIOT-OS/riotdocker/pull/109#issuecomment-701505405)
Instead of trying to sort that out by ourself, just use the package [provided by Debian](https://packages.debian.org/search?suite=bullseye&searchon=names&keywords=picolibc). It has ~~no dependencies~~ the respective gcc packages as dependencies. 

So instead of manually downloading the toolchain, also install that from the repository.


kaspar030:
didn't upstream miss risc-v support?

benpicco:
I've been compile-testing for the `hifive1` with it the whole time.

https://packages.debian.org/testing/picolibc-riscv64-unknown-elf

kaspar030:
somehow this failed *all* builds on travis.

benpicco:
Yes because I was wrong about the packages having no dependencies.
They depend on the respective gcc package (e.g. https://packages.debian.org/bullseye/gcc-riscv64-unknown-elf) which we also compile from source.

Dockerfile
```diff
@@ -162,6 +161,18 @@ RUN mkdir -p /opt && \
 
 ENV PATH $PATH:/opt/gnu-mcu-eclipse/riscv-none-gcc/${RISCV_VERSION}-${RISCV_BUILD}/bin
 
+# Install picolibc Debian package
+ARG PICOLIBC_VERSION=1.4.6-1
```
kaspar030, line 165: there's a 1.7.3 package, would that work, too?

benpicco: of course!

kaspar030:
@benpicco should we push this? how can I help?

benpicco:
Ah the problem is still that we install the compilers from outside the package manager, so the dependency of the picolibc package is not met.

For ARM, the 2019q4 version of GCC 9 we install manually is also available [on focal](https://packages.ubuntu.com/focal/gcc-arm-none-eabi), but for RISC-V our GCC 10.1.0-1.1 is a bit newer than what [focal has in store](https://packages.ubuntu.com/focal/gcc-riscv64-unknown-elf).

benpicco:
Looks like it worked this time!

benpicco:
bors try

kaspar030:
ok, but we're switching to the ubuntu gcc packages completely. do we want that?

benpicco:
bors try

kaspar030:
I don't really like that this switches to ubuntu's version of arm gcc :/

benpicco:
What's wrong with it?

kaspar030:
> What's wrong with it?

how has it been compiled? we've been using the arm toolchain for years. and the ubuntu one used to have quirks (I remember the small enum thing? or was that debian?).

benpicco:
I don't know, I've always just used the toolchain provided in the Ubuntu repository.

benpicco:
I would expect that most users are installing the toolchain from their distribution's repository as it's customary on Linux.

I didn't have any issues with using it for the past years.

Mind you that the source we are pulling from is [deprecated](https://developer.arm.com/downloads/-/gnu-rm) and if you want to upgrade to a [new version](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads) we'll have to adapt the build scripts.

I'm not keen on doing this tedious work when we could as well just install it from the repo instead.

maribu:
If there are indeed issues with the Ubuntu toolchain, it would be IMO in our best interest to address them by getting fixes upstream into the Ubuntu repos. Ubuntu is very much the most popular Linux distro and Ubuntu user will use the toolchain that is packaged for Ubuntu. We have to deal with this in any case.

For me, it would also be easier to be able to reproduce the issues Ubuntu users have by just passing `BUILD_IN_DOCKER=1` instead of having to set up a container / VM with Ubuntu. And ideally I would like to run into the issues myself during the CI run, rather than having someone else run into it.

maribu (approved):

kaspar030:
@maribu your call when to merge this (to not change riotdocker during release)

kaspar030:
bors merge

maribu:
bors merge

kaspar030:
bors ping

benpicco:
Thank you!

PR #118, opened 2020-10-02
requirements.txt: add colorama

benpicco:
The suit-manifest-generator script uses the [colorama](https://github.com/ARMmbed/suit-manifest-generator/blob/master/setup.py#L57) package.

In order to swich to the upstream version, also include it in our requirements.txt.

Needed for https://github.com/RIOT-OS/RIOT/pull/15095


kaspar030 (approved):
ACK.

PR #125, opened 2020-12-17
Dockerfile: add uncrustify

miri64:
In https://github.com/RIOT-OS/RIOT/pull/15667 I noticed that `uncrustify` is not run in the static tests, since it isn't installed into our Docker. Here is the fix for that.


maribu (approved):
ACK

PR #126, opened 2021-01-13
Dockerfile: add shellcheck

miri64:
See https://github.com/RIOT-OS/RIOT/pull/15756/files#r556615932


jia200x (approved):
ACK. I confirm this is the right apt package (see https://github.com/koalaman/shellcheck#installing).

miri64:
Thanks for the quick review!

PR #127, opened 2021-01-14
Refactor Docker images hierarchy

aabadie:
This PR is a proposal to rework the hierarchy of docker images within this repository.

The current state is to put everything (static test tools, toolchains, development helpers) in a single image `riot/riotbuild`. While this has advantages (simplicity and everything in the same place), this also has drawbacks: the resulting image is huge (+6GB) and contains things not necessarily needed depending on the use case (running static tests doesn't require the toolchains to be installed). I propose to split this image in several images, to allow sharable and more focused features.

The proposed approach is to split the `riot/riotbuild` image in 3:
- `riot/riotdocker-base` will be used as a base for all other images and contains things that are in common: git configuration, build the createuser tool and the `run.sh` entrypoint. It also configures the workdir.
- `riot/static-test-tools` is based on `riot/riotdocker-base` and install only the tools required by the static tests. This image could be used with the related CI checks in the RIOT repo (static-tests GH action and doc-build). The resulting `riot/static-test-tools` takes 660MB.
- `riot/riotbuid` is based on `riot/static-test-tools` so the resulting image contains the same features as in master and size is equivalent. In the long term, I plan to base this image directly on `riot/riotdocker-base`, once the CI checks in the RIOT repo are updated and use `riot/static-test-tools` directly.

If we agree with this PR, there are following steps needed:
- on the dockerhub side, someone must create new repositories for the new images, within the riot organization (I don't the rights for that)
- automate the build/test/publish of images when an update is pushed to the master branch. I'm not sure dockerhub handles nicely the dependencies between images with Dockerfiles in the same repo, that's why I think we have to do that. If someone thinks it's possible, then there's no need to do that of course. Also I found [this action](https://github.com/marketplace/actions/build-and-push-docker-images) could be useful, especially if we want to also publish to the github registry and get better performances with CI checks on the RIOT repo.


aabadie:
The new images on Dockerhub are created: [riot/riotdocker-base](https://hub.docker.com/repository/docker/riot/riotdocker-base) and [riot/static-test-tools](https://hub.docker.com/repository/docker/riot/static-test-tools).

With some configuration of the image builds on the Dockerhub side (path to the Dockerfile of the riotbuild image), I could make the Docker cloud check to succeed. If we agree on this PR and are about to merge it, ping me before merging, so I (or someone else) have time to update the build configuration on Dockerhub before, otherwise the automatic build of master will break.

riotbuild/Dockerfile
```diff
@@ -13,22 +13,19 @@
...
-# 1. Build own image based on latest base OS image:
+# 1. Build own image based on latest base OS image (from the riotbuild directory):
 #      # docker build --pull -t riotbuild .
 #
 # Usage:
 # 1. cd to riot root
 # 2. # docker run -i -t -u $UID -v $(pwd):/data/riotbuild riotbuild ./dist/tools/compile_test/compile_test.py
 
-FROM ubuntu:bionic
+FROM riot/static-test-tools:latest
```
aabadie, line 23: We could also directly use `riot/riotdocker-base` here. I don't have a strong opinion here, the current approach results in the same `riot/riotbuild` image. Thoughts ?

kaspar030: I think it needs to be this way for a while.

kaspar030 (approved):
ACK.

kaspar030:
Does this still honor the build hooks?

```
❯ dwqc -q breeze "cat /etc/riotbuild"
RIOTBUILD_VERSION=unknown
RIOTBUILD_COMMIT=unknown
RIOTBUILD_BRANCH=unknown

aabadie:
[yes](https://github.com/RIOT-OS/riotdocker/blob/5b07e4b14c3a7094eff3fd21ce65b0b66a092463/riotbuild/Dockerfile#L261-L274)

kaspar030:
> But where are the build hooks configured ?

there's the `hooks/` folder that docker hub uses. Maybe moving that to `riotbuild/` would do the trick?

kaspar030:
> there's the `hooks/` folder that docker hub ~uses~ used. Maybe moving that to `riotbuild/` would do the trick?

seems like, from [the docs](https://docs.docker.com/docker-hub/builds/advanced/#custom-build-phase-hooks):
"Create a folder called hooks in your source code repository _at the same directory level as your Dockerfile_. Place files that define the hooks in that folder."

I'll open a PR.

PR #129, opened 2021-02-02
riotbuild: include rapidjson extras for riotctrl

miri64:
The optional `rapidjson` dependency was added to `riotctrl` in https://github.com/RIOT-OS/riotctrl/pull/8 for a more lax parsing of JSON objects such as trailing commas within an object like this:

```json
{
  "one": 1,
  "two": 2,
  "three": {"I": "have", "more": "members",},
}
```

This adds that optional dependency. However, [dependency resolution was broken in `v0.2.0` and `v0.2.1`](https://github.com/RIOT-OS/riotctrl/pull/14), we need to use at least `v0.2.2`.

### DIFF
--- a/riotbuild/requirements.txt
+++ b/riotbuild/requirements.txt
@@ -15,5 +15,5 @@ scikit-learn==0.22.1
 joblib==0.14.0
 emlearn==0.10.1
 jinja2==2.11.2
-riotctrl>=0.1.1
+riotctrl[rapidjson]>=0.2.2
 pyhsslms>=1.0.0


MrKevinWeiss:
Who would have thought so many things need to be touched...

miri64:
> Who would have thought so many things need to be touched...

Just this one thing ;-).

MrKevinWeiss (approved):
Looks good.

I tested by importing rapidjson in python3 in the docker image without errors.

miri64:
Mhh docker cloud fails, but the link is a 404 :-/

miri64:
Let's see, if a force push solves this.

miri64:
Nope :(

miri64:
Another try.

miri64:
Ok, now even `riotdocker-base` fails, which does not include the change... this seems to be unrelated...

PR #141, opened 2021-04-02
riotbuild: Install Rust

chrysn:
I'm not yet sure this is all the right way to do things, but a) hope that you can tell me, and b) ask the CI to build this (for with my current uplink this all would take ages)

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -302,5 +302,7 @@ RUN echo "RIOTBUILD_VERSION=$RIOTBUILD_VERSION" > /etc/riotbuild
 RUN echo "RIOTBUILD_COMMIT=$RIOTBUILD_COMMIT" >> /etc/riotbuild
 RUN echo "RIOTBUILD_BRANCH=$RIOTBUILD_BRANCH" >> /etc/riotbuild
 
+ARG DOCKER_REGISTRY
+FROM ${DOCKER_REGISTRY}/c2rust-built:latest as c2rust-built
 # These depend on llvm-7-dev to be installed
-COPY --from=chrysn/c2rust-built /usr/bin/c2rust /usr/bin/c2rust-refactor /usr/bin/c2rust-transpile /usr/bin/
+COPY --from=c2rust-built /usr/bin/c2rust /usr/bin/c2rust-refactor /usr/bin/c2rust-transpile /usr/bin/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,9 @@ on:
   pull_request:
     branches:
       - '*'
+  workflow_run:
+    workflows: c2rust-build
+    types: completed
 
 jobs:
   build-test:
@@ -42,6 +45,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+
+      # If this was triggered by c2rust-build, ensure it is done by now.
+      - name: Workflow Run Wait
+        uses: ahmadnassri/action-workflow-run-wait@v1
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
--- a/.github/workflows/c2rust-build.yml
+++ b/.github/workflows/c2rust-build.yml
@@ -1,0 +1,48 @@
+name: c2rust-build
+
+on:
+  push:
+    paths:
+      - 'c2rust-built/**'
+    branches:
+      - master
+  pull_request:
+    branches:
+      - '*'
+    paths:
+      - 'c2rust-built/*'
+
+jobs:
+  build-c2rust:
+    runs-on: ubuntu-latest
+    name: Build C2Rust
+    env:
+      DOCKER_REGISTRY: "${{ secrets.DOCKER_REGISTRY || 'local' }}"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver: docker
+
+      - name: Build C2Rust
+        uses: docker/build-push-action@v2
+        with:
+          context: ./c2rust-built
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/c2rust-built:latest
+
+      - name: Login to DockerHub
+        if: "${{ github.ref == 'refs/heads/master' }}"
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Push Images
+        if: "${{ github.ref == 'refs/heads/master' }}"
+        run: |
+          docker image push ${{ env.DOCKER_REGISTRY }}/c2rust:latest
--- a/c2rust-built/README.md
+++ b/c2rust-built/README.md
@@ -2,14 +2,10 @@ This image contains built and usable versions of [c2rust].
 It is provided because upstream [does not release binaries] for c2rust and its companion tools,
 and because the [branch this is built from (for-riot)] contains some fixes to c2rust required to work on RIOT.
 
-As this changes rarely,
-and because on the github-workers infrastructure this is [difficult to get right],
-this is not built and used through CI yet,
-but manually built an published as:
-
-    docker build . -t chrysn/c2rust-built
+This is built as part of the local CI
+and published as riot/c2rust-built,
+but used mainly through how it is copied into the riotbuild container.
 
 [c2rust]: https://github.com/immunant/c2rust
 [does not release binaries]: https://github.com/immunant/c2rust/issues/326
 [branch this is built from (for-riot)]: https://github.com/chrysn-pull-requests/c2rust/tree/for-riot
-[difficult to get right]: https://github.com/RIOT-OS/riotdocker/pull/141


chrysn:
I'm approaching the end of my ideas of how to make this build; the `Compiling cargo v0.44.1` ran even into a 30' timeout. Maybe that's really too short as I had to put the `travis_wait` command around all of the docker build, trying once with a 45' timeout and gonna stop wasting build time if even that doesn't work out.

chrysn:
This really *is* a timeout issue; I now have a machine where I can run this without timeouts, and it does terminate (albeit with an error I'm just tracking further).

On the more general side, are we OK with increasing the build time by as much as this (don't have numbers, but this might double the build times), with hacks to extend the timeout? (Single Cargo crate compilation is simply nothing too verbose, and the `travis_wait` doesn't quite cut it; a du-based custom watcher (like "have a background process that prints directory sizes every five minutes; as long as it's still going that will increase") could do, but will we stay on travis anyway? I don't know much about that tool as I'm using GitLab for my projects, but the builds show a deprecation / migration warning.)

chrysn:
Hm, seems I had the wrong travis log open; the current version builds locally, and the latest build before this push (https://travis-ci.org/github/RIOT-OS/riotdocker/builds/765744396) indicates the timeout extension does work.

So if things go green here, the main question to remain is whether we want to have the increased build time and image size; I think the benefits of running Rust on RIOT in CI are worth it, but I'm obviously biased ;-)

chrysn:
… and rebased with the time-saving do-not-merge commit removed because now the builds run through the Docker image creation, and all steps in there are required to make the later hello-world build tests run through.

kaspar030:
> So if things go green here, the main question to remain is whether we want to have the increased build time and image size; I think the benefits of running Rust on RIOT in CI are worth it, but I'm obviously biased ;-)

I might be similarly biased, but having Rust is definitely worth a lot. I think we can cover up the increased image size ("giant" will still be "giant" afterwards), but, can we mitigate the build time by building c2rust externally?

c2rust should end up being a single binary, with some runtime dependencies (libllvm, ...). If we carve out c2rusts building into it's own Docker container, we can rebuild that only if c2rust changes, and then re-use the artifact for the build of riotdocker.
Maybe move that into another repository?

chrysn:
I've had https://github.com/immunant/c2rust/issues/326 open for some time on that topic. Options seem to be:

* Have a Docker build from the same image (we're on bionic) that can be merged in
* Have generally usable Linux binaries produced by c2rust that can just be wgot and installed by placing them into the container.

I have a personal preference for the latter, and am exploring that road; I'd welcome anyone familiar with Docker workings and customs to explore the other road.

chrysn:
Hm, they *do* already provide a docker image; that route may actually be easier even without good familiartiy with Docker. Playing around...

chrysn:
Next iteration: Turns out the image provided by immunant only contains everything required for building and testing c2rust and not c2rust itself. However, that makes it rather straightforward to create one where it's built too, and (until my two pending PRs there are merged) we need a customized build anyway.

The current setup uses one intermediate image (also built in here by Travis) based on the immunant image that just builds "our" c2rust; that image is added as a dependency of the riotbuild image and the single c2rust binary is copied over. It ran through locally, so if anything runs afoul I may need to introduce `travis_wait`s again…

chrysn:
With the latest fixes, this can now be now tested comprehensively by running, inside the produced riotdocker (at least with what I'm producing locally):

```
# git clone https://github.com/chrysn-pull-requests/RIOT.git -b rust-application
# cd RIOT/examples/rust-hello-world/
# make all term
```

(I assume it doesn't make much sense to include a test in the code at this stage where https://github.com/RIOT-OS/RIOT/pull/16274/ is not merged yet; testing the fundamentals manually, then testing that PR based on Rust support in riotdocker, and at some point later maybe adding the rust-hello-world example to the make tests.)

---

With the Travis Docker images, I'm still in trouble: There's a travis run for building the `c2rust-built` docker image now, and riotdocker depends on that, but somehow fails to pull it in unlike the static-test-tools one built just the same way a line above. I tried with https://github.com/RIOT-OS/riotdocker/pull/141/commits/49c332606d5e590157a8701ad824bacaec74d867 but that wasn't it -- any clues?

kaspar030:
> With the Travis Docker images, I'm still in trouble: There's a travis run for building the `c2rust-built` docker image now, and riotdocker depends on that, but somehow fails to pull it in unlike the static-test-tools one built just the same way a line above. I tried with [49c3326](https://github.com/RIOT-OS/riotdocker/commit/49c332606d5e590157a8701ad824bacaec74d867) but that wasn't it -- any clues?

could it be that because of the `--pull` in `docker build --pull -t riotdocker riotbuild`the riotbuild run *always* tries to pull from docker hub, and this just works for the other images because there's a version there (pushed earlier)? This would mean the current workflow is broken. @aabadie do you have an idea?

I think the `--pull` should not be used for riotbuild, but the base images.

chrysn:
All running fine through the changed parts up to the point when the global timeout reaps the job. What can we do about that? Build our c2rust in a separate repo? Or just build and upload one (it's a 2-liner Dockerfile) out of CI and publish that?

kaspar030:
bors try

kaspar030:
bors ping

chrysn:
Hm, bors runs into the same timeout. It's about 20' for the C2Rust compilation step, and in total this exceeds the 50' limit.

Splitting it up in different repos would be the obvious option, but that'll mean that changes in C2Rust build repo won't be CI-tested with the rest until they're merged and published. Then again, the risk is from breaking changes down at C2Rust anyway, and they don't CI with us (how should they with every downstream), so meh. I'm preparing such a step by publishing what's now in c2rust-built as chrysn/c2rust-built in preparation of the same image being tagged riot/c2rust-built.

Generally speaking, how *do* we publish riotdocker images? Is there an automatism from it being merged, or does someone locally build and publish them at regular intervals?

chrysn:
Runs through now. May I rebase?

chrysn:
I've taken the liberty to rebase this, given that #148 made this unmergable on its own. Some of the travis workarounds could vanish with that.

I'm a tad confused as to where the information about what to build actually went, hoping to get a bit of clarity from looking at the CI output.

chrysn:
I don't know much about the github workflows involved here, and have taken the recipes from the other docker containers in something between cargo-culting and guessing-what's-needed -- relying on reviewers to judge whether what's in there makes sense.

Having the container rebuilt on every CI run seems extraneous, can we somehow make it so that "Build c2rust" is only ever executed when c2rust/Dockerfile changes? (As that clones from a git branch, if it helps, I can make that clone a named revision, to ensure that the built container is as much a function of the Dockerfile as possible, and thus use Docker's memoization).

chrysn:
For documentation of why I'm pushing around so much, and where I'm stuck: The riotbuild stage is outputting errors like

```
error: failed to solve: failed to solve with frontend dockerfile.v0: failed to create LLB definition: circular dependency detected on stage: c2rust-built
Error: buildx failed with: error: failed to solve: failed to solve with frontend dockerfile.v0: failed to create LLB definition: circular dependency detected on stage: c2rust-build
```

which so far there are two search engine hits, neither particularly helpful. I've even tried breaking any cycles by not referring to the just-built c2rust container any more, but to one I published manually. Help would be appreciated.

chrysn:
bors try

(to see whether I can wield this magic too)

chrysn:
OK, that failed in the expected mode; we'll need to publish a riot/c2rust-built once either manually, or just change that before merging to s/local/c2rust/, and then bors can do the final tests and merge after they pass.

kaspar030:
> To recap, open question:

Can't help with those. I agree that we hopefully won't rebuild c2rust too often.

Overall this PR looks good. One thing that bugs me though: can we pin the nightly version somehow?
I know riotdocker is not really reproducible, but installing rust/nightly will result in different rust versions daily.

chrysn:
On Tue, Aug 31, 2021 at 01:50:32AM -0700, Kaspar Schleiser wrote:
> Can't help with those. I agree that we hopefully won't rebuild c2rust too often.

The options I see are:

* Somehow make github-actions skip the step most of the time (I'd need
  help with that, so probably no)

* Move c2rust-built into a separte repo parallel to riotbuild. That'd
  need some setting up and parallel procedures etc to get the container
  to be riot/c2rust-built

* Use it like any other non-RIOT-managed container: I'd keep the builder
  in the tree here, but not build it from CI, and publish it manually as
  chrysn/c2rust-built -- and the riotdocker container would just import
  from there.

  I kind of dislike this on grounds of pulling something more unmanaged
  into something more managed, but then again, we're probably pulling in
  all kinds of stuff into the containers from sources with different
  management flows, so maybe its' not that bad.

Opionions on these?

> Overall this PR looks good. One thing that bugs me though: can we pin the nightly version somehow?
> I know riotdocker is not really reproducible, but installing rust/nightly will result in different rust versions daily.

Yes I'll do that -- it'll cause occasional "bump nightly version" PRs
when any new features are used, but they're easy to ack and do make the
transitions more visible. Expect follow-up.

riotbuild/Dockerfile
```diff
@@ -272,3 +296,20 @@ ENV RIOTBUILD_BRANCH $RIOTBUILD_BRANCH
...
+# frontend dockerfile.v0: failed to create LLB definition: failed to parse
+# stage name "${DOCKER_REGISTRY}/c2rust-built": invalid reference format:
+# repository name must be lowercase
+#
+# Until github supports this, the hardcoded "local" works well enough for the
+# tests, and needs to be replaced manually with "riot" when going live. Once
+# replaced and published, the ill-effect downstream users will experience is
+# that if they change their c2rust build, they'll have to adjust this line.
+# (For others, c2rust will be built but not used here).
+COPY --from=local/c2rust-built /usr/bin/c2rust /usr/bin/c2rust-refactor /usr/bin/c2rust-transpile /usr/bin/
```
fjmolinas, line 315: When using multistage COPY, FROM etc... you just need to redeclare the ARG see https://github.com/fjmolinas/riotdocker/blob/master/flashers/edbg.Dockerfile as an example

chrysn: Thanks, I'll give it a try with the next CI test run.

fjmolinas: Try it locally first otherwise!

chrysn: buildx is not in Debian yet, and I'm very careful about third party debs, but I'm running all tests I can without it here, and with the reference script and the ARG it's hopefully just one more run.

chrysn: Hm, even with the repeated ARG I get the same circular dependency error; any clues?

(Then again, going for a different workflow probably sidesteps the issue -- so maybe it's best to let this be and go for the other workflow right away).

chrysn:
Converting this to a separate workflow, there's one thing I don't see how to do: How would I get the information about whether the c2rust workflow was run into the riotbuild/Dockerfile?

If the c2rust workflow was run, I'd need to set up the copy from as ${DOCKER_REGISTRY}/c2rust-built:latest (do I need to get any files around for that, or are they all in the runner's local environment?) -- but if the c2rust workflow was not run, it'd need to copy from the latest riot/c2rust-built.

chrysn:
Frustrated with the github worker setup, let's take this a step at a time -- with c2rust-built from a third party source (me). Let's focus on #152 first -- that's basically this but stripped down to treating c2rust-built as a third party image.

Once we're done with that (I'd guess it's pretty ready), we can resume work here to explore how to build a riot/c2rust-built on this infrastructure.

chrysn:
Given that the split solution of #152 Kind Of Works, I'm closing this for now; next steps are updating the c2rust build process anyway (#153) which simplifies how that (manually built / managed by me but still having its sources here) process works.

What remains (remained?) valuable from this PR is the machinery of building c2rust in here too; I don't see that happen right now (also for lack of pressure). As the branch in this PR is hard to read (needs diffing to #152 as that was squashed), I'm force-pushing a latest state here based on #152 that loses history but in return is understandable in terms of what is missing compared to there.

PR #142, opened 2021-05-19
.github: add workflow to build and test images, using `docker` driver

fjmolinas:
This PR is a simplified alternative to https://github.com/RIOT-OS/riotdocker/pull/136, the major difference comes from it using the `docker` driver instead of `docker-container` driver for buildx which basically makes it behave like regular docker, so: no-cache, no multiplatform build, but the workflow can be simplified a lot.

I personally think we lose a lot with this, but being able to replace Travis, Docker with only GitHub actions, integrate versioning is a huge plus and this PR should be easier to integrate and to understand (there are a lot of workarounds to be done with `docker-container` driver as seen in #136).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,131 @@
+---
+# This workflow will run on all pull-request branches and on 'master',
+# 'staging', and 'trying'
+#
+# In all cases it will incrementally build: riotdocker-base, static-test-tools
+# and riotbuild
+#
+# If riotbuild is successfully built then it will run a simple compile test on
+# a subset of BOARDs using gcc and llvm
+#
+# On master if the building and testing succeeds it will push the generated image
+# to ${DOCKER_REGISTRY}
+#
+# * Requirements:
+#   * Add the following secrets:
+#     * DOCKER_REGISTRY: the registry to pull and push images from/to
+#     * DOCKERHUB_USERNAME: the Docker Hub username account that can
+#       publish to ${DOCKER_REGISTRY}
+#     * DOCKERHUB_PASSWORD: the Docker Hub password for ${DOCKERHUB_USERNAME}
+#
+name: build
+
+on:
+  push:
+    branches:
+      - trying
+      - staging
+      - master
+  pull_request:
+    branches:
+      - '*'
+jobs:
+  build-test:
+    name: Build and Test
+    runs-on: ubuntu-latest
+    env:
+      RIOT_BRANCH: '2021.01'
+      DOCKER_REGISTRY: "${{ secrets.DOCKER_REGISTRY || 'local' }}"
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+        with:
+          driver: docker
+
+      - name: Build riotdocker-base
+        uses: docker/build-push-action@v2
+        with:
+          context: ./riotdocker-base
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/riotdocker-base:latest
+            ${{ env.DOCKER_REGISTRY }}/riotdocker-base:${{ env.RIOT_BRANCH }}
+
+      - name: Build static-test-tools
+        uses: docker/build-push-action@v2
+        with:
+          context: ./static-test-tools
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/static-test-tools:latest
+            ${{ env.DOCKER_REGISTRY }}/static-test-tools:${{ env.RIOT_BRANCH }}
+          build-args: |
+            DOCKER_REGISTRY=local
+
+      - name: Set environment variables
+        run: |
+          echo "RIOTBUILD_BRANCH=$(git rev-parse --abbrev-ref HEAD)" >> $GITHUB_ENV
+          echo "RIOTBUILD_COMMIT=$(git rev-parse HEAD)" >> $GITHUB_ENV
+          echo "RIOTBUILD_VERSION=$(git describe --always)" >> $GITHUB_ENV
+
+      - name: Build riotbuild
+        uses: docker/build-push-action@v2
+        with:
+          context: ./riotbuild
+          tags: |
+            ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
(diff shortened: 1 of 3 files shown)


fjmolinas:
bors try

fjmolinas:
Succeeded the build https://github.com/RIOT-OS/riotdocker/actions/runs/856350545 https://github.com/RIOT-OS/riotdocker/actions/runs/856328329.

I would propose merging this PR as is, this would publish an image to `riotci`, @kaspar030 can then test it. If that is OK, I would create a follow up PR that would remove the travis action, docker integration and change bors status check. Does this sound ok?

fjmolinas:
ping @kaspar030!

fjmolinas:
Maybe @aabadie would be interested as well? Or @miri64 ?

.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Login to DockerHub
+        if: "${{ github.ref == 'refs/heads/master' }}"
+        uses: docker/login-action@v1
+        with:
+          username: ${{ env.DOCKER_REGISTRY }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Push Images
+        if: "${{ github.ref == 'refs/heads/master' }}"
+        run: |
```
kaspar030, line 109: So RIOT_BRANCH needs to be set to the current devel branch, right?
And we'd probably need per-release branches as well?

fjmolinas: This is regarding where the GitHub image was triggered though, this had to do with if the workflow is triggered from master or from a PR

.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+            local/riotdocker-base:latest
+            ${{ env.DOCKER_REGISTRY }}/riotdocker-base:latest
+            ${{ env.DOCKER_REGISTRY }}/riotdocker-base:${{ env.RIOT_BRANCH }}
+
+      - name: Build static-test-tools
+        uses: docker/build-push-action@v2
+        with:
+          context: ./static-test-tools
+          tags: |
+            local/static-test-tools:latest
```
aabadie, line 44: This line doesn't seem required since you set `DOCKERHUB_REGISTRY=local` in `build_args`.

BTW, why not use the same `build_args` strategy with the `riotdocker-base` image build ?

fjmolinas: It's not the same, one is used to tag the generated image, the other is used to define the base image for that stage build, so the `FROM` statement in the dockerfile

aabadie: Hmm, ok. So I think I don't understand why the `local/` is needed and why this env is required:

```
DOCKER_REGISTRY: "${{ secrets.DOCKERHUB_USERNAME || 'local' }}"
```

.github/workflows/build.yml
fjmolinas: Ah I understand your comment better, so currently, the workflow was not taking into account backports. In that case it would be indeed better to have a branch for the release. But will that be needed?

.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+          ref: ${{ env.RIOT_BRANCH }}-branch
+          path: RIOT
+
+      - name: GNU build test
+        run: |
+          make -CRIOT/examples/hello-world buildtest
+        env:
+          BUILD_IN_DOCKER: 1
+          DOCKER_IMAGE: local/riotbuild:latest
+          BOARDS: "arduino-uno esp32-wroom-32 hifive1 msb-430h native pic32-wifire samr21-xpro"
```
aabadie, line 83:
```suggestion
          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native pic32-wifire samr21-xpro"
```

Can we also add one atxmega and one esp8266 based boards ?

.github/workflows/build.yml
fjmolinas: When building on a PR, secrets are not passed, so `secrets.DOCKERHUB_USERNAME` does not exist.

aabadie: So, DOCKER_REGISTRY is set to local in this case. So there's no need for this line I guess.

.github/workflows/build.yml
fjmolinas: sure

.github/workflows/build.yml
fjmolinas: Ahhh yes seems like it

.github/workflows/build.yml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Build riotbuild
+        uses: docker/build-push-action@v2
+        with:
+          context: ./riotbuild
+          tags: |
+            local/riotbuild:latest
+            ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
+            ${{ env.DOCKER_REGISTRY }}/riotbuild:${{ env.RIOT_BRANCH }}
+          build-args: |
+            DOCKERHUB_REGISTRY=local
```
kaspar030, line 65: is it ok that sometimes it is DOCKER_REGISTRY and sometimes DOCKERHUB_REGISTRY?

fjmolinas: Unified to `DOCKER_REGISTRY`

.github/workflows/build.yml
fjmolinas: Added

.github/workflows/build.yml
fjmolinas: Done

fjmolinas:
27e2c85 fixes an introduced newline in BOARDS

.github/workflows/build.yml
fjmolinas: For some reason the build failed when I did, maybe it was a yaml syntax issue. I'll add the extra BOARDs on a follow up, currentyl I stick to mimic what is done in master.

fjmolinas:
@kaspar030 the action succeeded can I squash?

kaspar030:
yes, please squash!

kaspar030:
bors merge

fjmolinas:
bors merge

kaspar030:
bors merge

kaspar030:
bors try

kaspar030:
bors try

kaspar030:
ACK.

bors merge

PR #143, opened 2021-06-03
make travis fail on error exit of any step (`set -e`)

kaspar030:
Some builds continue even though an early step fails,
e.g., https://travis-ci.org/github/RIOT-OS/riotdocker/builds/773366924.

This does a `set -e`. let's see if this fixes the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ env:
   - RIOT_BRANCH=2021.01-branch
 
 script:
+  - set -e
   - docker build -t riot/riotdocker-base riotdocker-base
   - docker build -t riot/static-test-tools static-test-tools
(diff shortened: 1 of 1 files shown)


kaspar030:
Seems to work in the error case: https://travis-ci.org/github/RIOT-OS/riotdocker/builds/773371202

aabadie:
Should we rather remove the .travis.yml file in #142 (since it does the same job) ?

kaspar030:
> Should we rather remove the .travis.yml file in #142 (since it does the same job) ?

both, but this single line change makes the failures go faster.

kaspar030:
bors try

kaspar030:
closing in favour of #142.

PR #144, opened 2021-06-03
travis: log into docker hub when building bors branches

kaspar030:
This PR makes travis log into docker hub.

This is only done when building the bors branches (master, trying, staging), as for travis PR branch builds, travis doesn't supply the necessary environment variables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,10 @@ env:
   - RIOT_BRANCH=2021.01-branch
 
 script:
+  # log in to docker hub for bors builds
+  # (travis doesn't pass the necessary env vars to PR builds)
+  - echo "staging trying master" | grep -qw "$TRAVIS_BRANCH" && echo "$DOCKERHUB_TOKEN" | docker login -u "$DOCKERHUB_USER" --password-stdin
+
   - docker build -t riot/riotdocker-base riotdocker-base
   - docker build -t riot/static-test-tools static-test-tools
   - docker build --pull -t riotdocker riotbuild 


kaspar030:
bors try

fjmolinas (approved):
ACK

kaspar030:
bors merge

aabadie:
Is it possible to configure bors to use github-actions, so with #142 ? Because it seems that both github actions and travis are duplicating CI actions.

kaspar030:
> Is it possible to configure bors to use github-actions, so with #142 ? Because it seems that both github actions and travis are duplicating CI actions.

yes

kaspar030:
here's the plan: https://github.com/RIOT-OS/riotdocker/pull/142#issuecomment-843953344

aabadie:
So why is this PR needed since in the end the travis configuration will be removed ?

aabadie:
> here's the plan: #142 (comment)

That sounds a bit counter productive. I would have done all this in #142 (much simpler, less CI runs, round-trips, etc). And no need for another CI account/token on dockerhub.

fjmolinas:
> So why is this PR needed since in the end the travis configuration will be removed ?

Because to merge we first need travis to pass, and currently it does not

aabadie:
> Because to merge we first need travis to pass, and currently it does not

Ok, but if in #142, the bors configuration is changed to point to github actions, would it work ? If #142 also removes the travis configuration that should be fine as well, since it's a replacement. But maybe I'm missing something in the way bors works.

fjmolinas:
> > Because to merge we first need travis to pass, and currently it does not
> 
> Ok, but if in #142, the bors configuration is changed to point to github actions, would it work ? If #142 also removes the travis configuration that should be fine as well, since it's a replacement. But maybe I'm missing something in the way bors works.

Yes it would. You are not missing anything except that #142 does not want to replace Tavis, it wants to see that we can do exactly the same thing, and the final step is for it to build master and push the generated images. And so I didn't want to replace the current CI (travis + dockerhub hooks) before that final stage was validated in master. Once thats validated then travis is indeed no longer needed.

PR #147, opened 2021-06-04
.github/workflows/build: login on trying

fjmolinas:
The action failed again at the login stage, setting this up as a fail fast.


fjmolinas:
bors try

PR #149, opened 2021-06-07
.github/workflows/build.yml: bump RIOT branch

fjmolinas:
I initially misunderstood how our branches were named, and though I could re-user the RIOT_BRANCH use for testing as the tag for the image, but IMO it makes more sense to tag the next release version. Instead, so how I see it would be:

- tag the DOCKER image with the next release version (so in current case `2021.07`) while building against `2021.04-branch`
- once release is over switch RIOT_BRANCH to `2021.07-branch` and `VERSION_TAG` to `2021.10`

Maybe it's a good time to discuss how to version our images as well.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,7 +34,8 @@ jobs:
     name: Build and Test
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


kaspar030:
ACK.

bors merge

PR #152, opened 2021-09-01
riotbuild: Install Rust (minimal edition)

chrysn:
The c2rust program (required in addition to Rust itself) is installed
for what is for most intents and purposes a third party container, but
happens to have its sources shipped along the other containers.

---

This is a minimal version competing with #141 -- we can still get c2rust to fully integrate over time, but right now there's a third party (happens to be me) who publishes a suitable base container, which is not much more or less transparently managed than any other base container (especially immunant/c2rust which it itself is based on), and hey by the way even the Dockerfile for that container is shipped here (but more for informational purposes, not because CI would access it).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,6 +29,7 @@ on:
   pull_request:
     branches:
       - '*'
+
 jobs:
   build-test:
     name: Build and Test
--- a/c2rust-built/Dockerfile
+++ b/c2rust-built/Dockerfile
@@ -1,0 +1,11 @@
+FROM immunant/c2rust:ubuntu-bionic-latest as c2rust
+
+# --locked --git --branch ignores --locked; new (2021-03-something) cargo
+# versions behave properly, but C2Rust is on a 2019-12-05 toolchain.
+#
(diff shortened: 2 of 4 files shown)


chrysn:
By the way, the container built from this is successfully in use for the [riot-examples](https://gitlab.com/etonomy/riot-examples/) CI (so successfully that I've started building more in CI than was previously tested, which is why it's red again).

kaspar030:
bors try

kaspar030:
Regular builds seem to still work, so lets go.

bors merge

kaspar030 (approved):

kaspar030:
Unfortunately due to Docker not doing auto builds anymore, this won't be used for Murdock until I move the murdock container builds somewhere else. That's high on my list...

chrysn:
Thanks, working to improve from here now. (Given that the trouble building on native with this comes from mismatched LLVM 7.0 and 7.1 (from c2rust-build), I expect best progress after #104.)

Is there anything I can do on the murdock side (being relatively unfamiliar with its inner workings)?

PR #153, opened 2021-09-02
Build C2Rust on focal

chrysn:
This issue tracks changes needed in the Rust installation (OK: C2Rust -- the rest should just work) when #104 is rebased and gets merged.

By building c2rust on the same Ubuntu release (focal), the friction between different installed LLVM versions will be reduced.

This
* switches from using c2rust's shipped build containers (which made sense back when supported Ubuntus had those old LLVM versions) to a plain feisty base,
* consequently installs the build-deps manually, and
* builds a Debian binary package in the course of installing c2rust.

The latter is done to get us rid of the manual `apt-get install llvm-7-dev` (and pray that it matches what c2rust needs) in favor of copying the .deb out of the c2rust image and getting the right build-deps automatically. (A part of building a Debian package is the tools inspecting any binaries for used shared libraries, looking up the packages they're in, and looking at that library's exported symbol versionings to get a correctly version for the dependency).

<del>Put on hold / WIP / draft until #104 is done, given that these don't need lockstepping yet.</del>

### DIFF
--- a/c2rust-built/Dockerfile
+++ b/c2rust-built/Dockerfile
@@ -65,7 +65,9 @@ RUN \
 #
 # # Cleaning up as the relevant files were copied out, and the rest would just
 # # make for a large image (which should only contain a minimal delta to the base
-# # image in the end)
+# # image in the end: the installed dependencies (libllvm), the installed
+# # binaries and the debs that contain the stripped binaries and separate debug
+# # symbols once more)
 # RUN \
     rm -rf c2rust && \
     rm -rf ~/.cargo && \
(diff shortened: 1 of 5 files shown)


chrysn:
With #104 done this is now ready for review.

The large changes have already been pushed to dockerhub and are thus available as chrysn/c2rust-built:focal; the only "real" change in this PR is that that is now actually used, which helps getting everyone agree on the same standard library files. (The other changes are also "real" but don't effect the builds directly, but indirectly because I uploaded :focal based on them already).

On a side note, c2rust-built is more and more becoming a Debian source package (but with no chance of upstreaming due to C2Rust's idiosyncrasies, I'm not going all the way of converting it, also because we have infrastructure for Docker builds but not for Debian builds).

kaspar030 (approved):
ACK.

bors merge

chrysn:
Thanks!

PR #156, opened 2021-10-19
integrate murdock worker

kaspar030:
The CI containers were previously built from https://github.com/kaspar030/riotdocker/tree/dwq, which is a fork of this repo, but completely changes its contents. Now the container is not built anymore due to docker hub changes.

This PR copies the dwq branch contents into a subfolder of this repo and adds building to the github workflow.

Wherever it made sense, I already changed "-slave" to "-worker". The scripts need adapting still, though.

*note* the push location is different from the current production one. The workers need to update their init script / systemd files to change the location.

Suggested testing:

1. if building of the image succeeds, we merge this, to test pushing of the resulting image
2. on success, I update one of the murdock workers to use the new image
3. if that works well, we can migrate the other workers
4. last, I'll remove the obsolete image to prevent anyone from using it.


fjmolinas:
> > Can you give this a test run in your own fork before we merge @kaspar030? You would only need to configure:
> 
> You mean the "push" step, those two lines, might fail?

Well, it is a few extra steps probably, but you would have a testing playground. But no I don't expect the push to fail. But since I'm nitpicking a couple of things I guess something could go wrong...

PR #158, opened 2021-11-02
riotbuild: use pycryptodome instead of pycrypto

miri64:
This replaces the [unmaintained `pycrypto`](https://github.com/pycrypto/pycrypto/issues/173) (which has several vulnerabilities, see https://github.com/RIOT-OS/RIOT/pull/17107) with the more up-to-date [`pycryptodome`](https://pypi.org/project/pycryptodome/).

### DIFF
--- a/riotbuild/requirements.txt
+++ b/riotbuild/requirements.txt
@@ -2,7 +2,7 @@
 pyasn1==0.4.2
 ecdsa==0.13.3
 pexpect==4.8.0
-pycrypto==2.6.1
+pycryptodome==3.11.0        # used by mcuboot
 pyserial==3.4
 ed25519==1.4
 cbor==1.0.0


miri64:
bors try

kaspar030:
should we *add* pycryptodome, instead of replacing pycrypto? just for a while?
that way we don't break anything that we didn't know uses pycrypto?

miri64:
> should we _add_ pycryptodome, instead of replacing pycrypto? just for a while?

That would not work. They have the same API so installing both would cause a conflict. See also https://www.pycryptodome.org/en/latest/src/installation.html. We could install `pycryptodomex` for the alternative API, but then that alternative API wouldn't be used by our tests.

miri64:
Also, I do not see much point in keeping an unmaintained security-related library with publicly known vulnerabilities.

fjmolinas (approved):
ACK

PR #159, opened 2021-11-05
riotbuild: bump ARM gcc to 10.3-2021.10

kaspar030:


kaspar030:
I don't think so. CI here does not run any binaries though. I can manually update one CI worker to this image for testing.
(not today though, queue is too long for experiments. :) )

fjmolinas:
Lets get this in after https://github.com/RIOT-OS/riotdocker/pull/180

kaspar030:
Well, there's now one test (tests/fmt_print) that gets run on nrf51 (microbit). IMO that should be sufficient, even with full release tests we wouldn't cover everything. And we don't expect the toolchain to produce broken binaries.

fjmolinas:
bors merge

PR #161, opened 2021-11-11
Rust updates

chrysn:
* Add what is needed to experiment with AVR (not fully supported for Rust-on-RIOT, but way easier to test being in the docker build as I haven't had AVR toolchains installed around for a long time, though they probably still can be installed easily)
* Update Rust nightly (that's a semi-regular thing since we decided to pin the version in https://github.com/RIOT-OS/riotdocker/pull/141#issuecomment-909035011 which also went into #152)

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -280,11 +280,16 @@ ENV PATH ${PATH}:/opt/rustup/.cargo/bin
 # toolchain add` or similar; CARGO_HOME is not because the user will need to
 # write there, and all rustup does here is to place some binaries that later
 # fan out to RUSTUP_HOME anyway.
+#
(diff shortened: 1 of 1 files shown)


chrysn:
bors try

(Is that the way you say it?)

chrysn:
Background: The new nightly makes [the RIOT examples CI](https://gitlab.com/etonomy/riot-examples) work again, which currently fails as the old nightlies don't understand the 2021 edition that crates start using.

On a local test of the examples with BUILD_IN_DOCKER=1 DOCKER_IMAGE=riotbuild:latest, they pass (at least the currently failing stage) again.

kaspar030 (approved):
ACK. shouldn't affect other builds, so lets go with it!

kaspar030:
bors merge

PR #162, opened 2021-11-17
riotbuild: Drop :focal tag in favor of a more stable one

chrysn:
This drops a workaround that was necessary during evaluation of #153 in
favor of a more long-term usable distinction between experimentation and
stable delivery, which is necessary until #141 is done properly and
c2rust is built in one go with the rest of this.

---

Cleanup change with no expected actual impact. (Tags diverge because `latest` = `for-riot` was rebuilt after #153 was merged, but any differences should be just because building is not byte-by-byte reproducible.)


kaspar030 (approved):
bors merge

PR #165, opened 2021-11-30
murdock-worker: bump git-cache

kaspar030:
bump git-cache version.

provides much faster tag handling.


MrKevinWeiss (approved):
With all our playing around seemed to indicate that it is an improvement.

ACK.

kaspar030:
bors merge

PR #166, opened 2021-11-30
riotbuild: bump riscv toolchain to 10.2

aabadie:
The minor fixes correspond to this warning raised by Docker: `apt does not have a stable CLI interface. Use with caution in scripts.`. So use apt-get instead which is consistent with the rest of the commands.

There's a small name duplication removed with coccinelle package using an ARG.

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -155,7 +155,7 @@ ENV MIPS_ELF_ROOT /opt/mips-mti-elf/${MIPS_VERSION}
 ENV PATH ${PATH}:${MIPS_ELF_ROOT}/bin
 
 # Install RISC-V binary toolchain
-ARG RISCV_VERSION=10.1.0-1.1
+ARG RISCV_VERSION=10.2.0-1.2
 RUN mkdir -p /opt && \
         wget -q https://github.com/xpack-dev-tools/riscv-none-embed-gcc-xpack/releases/download/v${RISCV_VERSION}/xpack-riscv-none-embed-gcc-${RISCV_VERSION}-linux-x64.tar.gz -O- \
         | tar -C /opt -xz && \


aabadie:
bors try

riotbuild/Dockerfile
```diff
@@ -151,7 +151,7 @@ ENV MIPS_ELF_ROOT /opt/mips-mti-elf/${MIPS_VERSION}
 ENV PATH ${PATH}:${MIPS_ELF_ROOT}/bin
 
 # Install RISC-V binary toolchain
-ARG RISCV_VERSION=10.1.0-1.1
+ARG RISCV_VERSION=10.2.0-1.2
```
fjmolinas, line 154: Unrelated?

fjmolinas (commented):
I don't think the RISCV bump belongs here, can you split that one out? Otherwise all good.

riotbuild/Dockerfile
aabadie: That's the goal of this PR :)

fjmolinas: Ahh sorry, I didn't read the PR title, only the description.

fjmolinas:
@aabadie I wonder if we should wait for the release before getting this in, although minor version bump issues could still surface...

kaspar030:
@aabadie would you please split out the minor fixes? Those we could merge anytime (I'd ACK them right away).

aabadie:
> would you please split out the minor fixes?

Done in #175

fjmolinas:
Lets get this in after https://github.com/RIOT-OS/riotdocker/pull/180

fjmolinas:
bors merge

fjmolinas (approved):
ACK

fjmolinas:
bors merge

PR #168, opened 2021-12-21
riotbuild: add qemu-system-arm

kaspar030:
This PR adds QEMU with it's full-system ARM emulator.

At least the microbit board can be emulated to a large degree with this.

A second commit adds a basic test run of `tests/fmt_print` for microbit, using qemu.

PR #169, opened 2021-12-22
test PR

kaspar030:
https://github.com/RIOT-OS/riotdocker/pull/168 confused me. testing whether docker actually picks up the newly built container.

DON'T MERGE.


kaspar030:
> testing whether docker actually picks up the newly built container.

confirmed. I think I got the issue in #168.

PR #170, opened 2022-01-07
Update Rust nightly version

chrysn:
With this version it will be possible to use cstr_core >= 0.2.5 (which
future versions of riot-wrappers will depend on)

Testing procedure: Things still build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -121,6 +121,22 @@ jobs:
           DOCKER_IMAGE: ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
           BOARDS: "native samr21-xpro"
 
+      - name: Rust build test
+        run: |
+          # Once 2022.01 is released, the switcharoo can be removed.
+          # Note that `git switch master` does not work because the checkout
+          # action does only minimal fetching.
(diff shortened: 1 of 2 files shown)


kaspar030:
Maybe it is time to add a simple Rust build to the workflow? One app, one board?

chrysn:
I thought the test involved running the RIOT CI anyway (even though not through murdock). Hm, only does static tests now -- and the three (GNU, GNU microbit qemu, LLVM) manual tests.

I'm currently looking at doing more Rust tests also with the riot-sys and riot-wrappers crate; would it be excessive to build a test on all boards in the riotdocker CI? (I wouldn't know which board to pick as the most likely problematic).

chrysn:
Added and passing through the new test (using the hello-world example and not bothering to run tests natively or emulated: experience tells that if something is wrong it likely won't build, and miscompilations or other runtime failures are so hard to spot if they happen that the simple test likely won't catch them either).

Some branch switching is introduced on the RIOT checkout; this can be removed later when a version with Rust support has been released.

kaspar030 (approved):
LGTM.

kaspar030:
bors merge

PR #172, opened 2022-01-14
murdock-worker: bump dwq to 0.0.51

kaspar030:
bumping dwq again.

0.0.51 has a small change, increasing the safety margin the worker uses to time-out jobs itself. This is to prevent a job from being timed-out by disque server, at which point it's output would be lost.


chrysn (approved):
Great, thanks, ACK.

Money quote:
```patch
-                        # be sure to timeout a bit earlier
-                        timeout -= 1
+                        # be sure to timeout a bit earlier, so transmit/network delays
+                        # don't make disque time-out itself.
+                        timeout -= 10
```

chrysn:
bors merge

PR #173, opened 2022-01-18
C2Rust: Use upstream as patches are not required any more

chrysn:
C2Rust has accepted most of the patches in the current for-riot; what's left needs riot-sys fixes anyway on the long run.

### Issues/PRs references

A sibling PR on RIOT docs is https://github.com/RIOT-OS/RIOT/pull/17536

### Status

This is on hold as a draft because it still depends on two things:

* [x] The no-llvm-asm branch of riot-wrappers to be merged, released and the Cargo.lock files here to be updated to use it, and
* [x] The unpatched C2Rust won't work for older Rust-on-RIOT versions (fixable with a `cargo update`, but that shouldn't be necessary for users to do to run the demos), so this will need to wait until the current release is through
 
-- and then maybe some time for good measure as to not cause any trouble for people on branches that haven't received the no-llvm-asm riot-wrappers update from RIOT yet.

[edit: Turned into a checklist, as first item is done]

### DIFF
--- a/c2rust-built/Dockerfile
+++ b/c2rust-built/Dockerfile
@@ -2,13 +2,17 @@ ARG DOCKER_REGISTRY="riot"
 FROM ${DOCKER_REGISTRY}/riotdocker-base:latest
 
 COPY debian/control /dpkg-build/debian/
+# When playing with these, it is often practical to remove the comments on the
+# `# true` / ... / `# RUN \` blocks (making the resulting images larger because
+# of the intermediate files in the chain, but quick to rebuild on small
+# changes) and moving the COPY lines down before the `dpkg-buildpackage` to not
+# download and install all the dependencies again. The control file needs to
+# stay up there as it influences what will be installed.
 COPY debian/rules /dpkg-build/debian/
 COPY debian/changelog /dpkg-build/debian/
+COPY debian/install /dpkg-build/debian/
 
 # noninteractive for the tzinfo
-#
-# The first block is for actually building c2rust, the second for building a
-# Debian package.
 RUN \
     echo 'Update the package index files to latest available versions' >&2 && \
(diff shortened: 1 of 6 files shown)


chrysn:
Also, given that it's now generic-C2Rust and not RIOT-vendored, it might be a good time to go all the way to making it a Debian package; it would still be buildable through the Dockerfile, but also in a standalone way.

chrysn:
The 2022.01 release is through, and I'd like to update this now -- one more step toward doing things the nice and easy way.

Along with the mere version and upstream source change, I've moved a few components over into the Debian package creation. This makes it easier to build a suitable c2rust also for other containers that might not be based on the precise Ubuntu version riotdocker is currently using, and the README reflects that.

chrysn:
Tested by building riotdocker on it, and building the rust-gcoap example for native, microbit-v2 and hifive1b.

(Not that I'd expect any change -- this is practically just switching to upstream which merged my PRs, and changing things about *how* c2rust is built).

kaspar030 (approved):
ACK.

(This is simpler now, less forks involved, but the c2rust build story is still something that I wouldn't read to my kids ;) )

kaspar030:
bors merge

chrysn:
Thanks, submission in process.

> c2rust build story

Fortunately, upstream has started porting it away from using compiler internals to using regularly released crates ([albeit still at an early stage](https://github.com/immunant/c2rust/issues/364)). Once that's done, it'll be a less-than-5-minutes bed time story ending with "and they used `cargo install c2rust` happily ever after".

kaspar030:
> Once that's done, it'll be a less-than-5-minutes bed time story ending with "and they used `cargo install c2rust` happily ever after".

nice, looking forward to that!

PR #177, opened 2022-02-04
murdock-worker: add "default-first" to MURDOCK_QUEUES

kaspar030:
updating MURDOCK_QUEUES default.

This has no effect to deployed workers, as that init file is put in place manually.


maribu (approved):

kaspar030:
bors merge

PR #182, opened 2022-02-11
actions: remove obsolete switcharoo

kaspar030:
As per the note, it should not be necessary anymore.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -123,12 +123,7 @@ jobs:
 
       - name: Rust build test
         run: |
-          # Once 2022.01 is released, the switcharoo can be removed.
-          # Note that `git switch master` does not work because the checkout
-          # action does only minimal fetching.
-          (cd RIOT && git fetch origin master && git checkout FETCH_HEAD)
           make -CRIOT/examples/rust-hello-world buildtest
-          (cd RIOT && git switch -)
         env:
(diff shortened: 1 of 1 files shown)


chrysn (changes requested):
Sadly it is not as obsolete as I had hoped, see notes on https://github.com/RIOT-OS/riotdocker/pull/186 where I (having missed the review-request for this one) tried the same and failed.

Practically, this will need to stay as long as nightly features used by the code still change from one Rust version to the other (which, I hope, will be after 1.61 -- the two nightly features left then are relatively stable in practice).

The build tests passed here -- but on the next nightly update they might (or would) fail again.

[edit: added last paragraph]

chrysn:
I think we should manage that after the 2022.10 release. There are, as of right now, still crates in the dependency tree that declare a nightly `#![feature(…)]`, but the features have been stabilized. While these can not be used with stable, nightly versions (and the snapshot version in 2022.10 will build them with nightly) will already have stable code for them.

Between 2022.10 and 2023.01 we will upgrade our stable Rust and could, in the course of that, remove nightly Rust from our images completely (see #208) -- if we do that, we'd need to wait another release cycle to merge this PR, as then the 2022.10 release would be missing a toolchain that can build its tests.

kaspar030:
Closed in favour of #221.

Issue #184, opened 2022-03-05
Confusing build error for `riotbuild`


gschorcht:
The installation of `emlearn` seems to fail in

https://github.com/RIOT-OS/riotdocker/blob/14fa5cda36f0f09282f3d1efdca79b7c1cb83dca/riotbuild/Dockerfile#L271-L273

because of missing `pybind11` module, but is actually successful. This is very confusing and I spent some time before I realized that building the wheel for `emlearn` fails only in first iteration but later succeeds. I just missed the small success message after the large number of red error messages.

The reason is that the wheels for `ed25519`, `cbor`, `scapy`, `emlearn` and `riotctrl` are built before the other collected packages are installed.
```
Collecting pybind11>=2.2
  Downloading pybind11-2.9.1-py2.py3-none-any.whl (211 kB)
...
Building wheels for collected packages: ed25519, cbor, scapy, emlearn, riotctrl
 ...
  Building wheel for emlearn (setup.py): started
  Building wheel for emlearn (setup.py): finished with status 'error'
  ERROR: Command errored out with exit status 1:
   command: /usr/bin/python3 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-wdbb97j4/emlearn/setup.py'"'"'; __file__='"'"'/tmp/pip-install-wdbb97j4/emlearn/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /tmp/pip-wheel-ev4akp51
       cwd: /tmp/pip-install-wdbb97j4/emlearn/
  Complete output (85 lines):
...
A LOT OF RED ERROR OUTPUT (85 lines)
...
ModuleNotFoundError: No module named 'pybind11'
  ----------------------------------------
  ERROR: Failed building wheel for emlearn
...
Running setup.py clean for emlearn
  Building wheel for riotctrl (setup.py): started
  Building wheel for riotctrl (setup.py): finished with status 'done'
  Created wheel for riotctrl: filename=riotctrl-0.5.0-py3-none-any.whl size=17456 sha256=8b12106d601e9a2c3592a5aeea656ce4c62e24e324895328e281e751c31c1395
[...]

PR #185, opened 2022-03-05
riotbuild: avoid errors when emlearn is installed

gschorcht:
As reported in issue #184, the installation of `emlearn` seems to fail in because of missing `pybind11` module, but is actually successful. The reason is that the wheels for `ed25519`, `cbor`, `scapy`, `emlearn` and `riotctrl` are built before the other collected packages are installed.

The extensive error output on the first failed iteration of the `emlearn` installation is very confusing and can be prevented by explicitly installing the `pybind` package before installing the other required packages, as suggested in this PR.


kaspar030 (approved):
ACK.

bors merge

PR #186, opened 2022-03-08
rust: Update nightly version

chrysn:
This ensures that after the 2020.04 release the supported stable version
can be bumped to Rust 1.61. Using Rust 1.61 to its full potential will
require removing feature declarations, and if riot-wrappers has those
features removed, building it needs a recent nightly (like this) that
does not need them declared.

In addition, it removes a workaround that was needed before the 2022.01 release for the build tests; they are now performed on whatever RIOT branch is selected (IIUC the last release -- anyhow, like all the other build tests).

---

In terms of testing, this should be covered well by the test; in addition, local build tests are being done against the latest RIOT master, and with a larger example than the hello-world.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -129,7 +129,11 @@ jobs:
 
       - name: Rust build test
         run: |
-          # Once 2022.01 is released, the switcharoo can be removed.
+          # Once a release is out that can be built on stable, this switcharoo
+          # can be removed -- until then, each release of RIOT is only
+          # supported with the very Rust version that was the configured
+          # nightly it was released with; such is the nature of nightly.
+          #
           # Note that `git switch master` does not work because the checkout
           # action does only minimal fetching.
           (cd RIOT && git fetch origin master && git checkout FETCH_HEAD)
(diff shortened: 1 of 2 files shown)


chrysn:
The second change was premature, I had to revert it; the latest Rust version will for some more time only support the master branch (with the nightly that was set at release time working for the last release).

The general issue are nightly features, which over the release cycles are becoming fewer, but with the versions of the last release we're not there yet.

The current version has the note on the switcharoo updated to more accurately represent the situation. Running on stable Rust will make things better; note though that while https://github.com/RIOT-OS/RIOT/issues/17526 (fundamentally work on stable) is about to be closed, there will still be platforms (riscv) and components (saul and shell until 1.61, coap-message probably longer) that continue to require nightly features.

kaspar030:
So, CI passes, this can go in, right?

chrysn:
In principle yes; if https://github.com/RIOT-OS/riotdocker/pull/173 finds approval soon too, it'd be better to merge that, then I'll upload the c2rust container, and then we can build/merge this to have both in one go. (As the c2rust build is not integrated but treated as a third-party docker image, merging the 173 would not trigger an actual update, but this will, and could do both in one go)

Suggestion: If you approve both, ACK this and merge the other, then I'll force this to test again and tell bors to merge after I've acted on the other.
If anything holds up the other, this can go in right away.

kaspar030 (approved):
ACK.

chrysn:
bors merge

now that the c2rust image has been updated. (In parallel, the build-and-test is running again after I triggered that, given that AIU the build tests are only part of that and not part of bors.)

PR #187, opened 2022-03-09
rust: Add beta version

chrysn:
With builds of RIOT on stable Rust becoming more and more attainable (basics are in, 1.60 is adding builds on native without undue complexity, 1.61 will add support for shell and SAUL), at some point we should make that part of the test suite.

The present commit is a first step there. It adds "beta", but the [plan](https://forum.riot-os.org/t/release-2022-04-dates-and-feature-requests/3548/3) is to switch that over to 'stable' when 1.60 becomes stable.

This likely doubles the size of the installed Rust (rustup itself is common but rather small; compiler, core libraries and source code will be duplicated); the build tests that give me numbers are still running. Possibly, the source code can be left out of the beta image -- that's mainly needed when rebuilding the core libraries to the target spec, which is available only on nightly.

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -293,17 +293,20 @@ RUN \
     RUSTUP_HOME=/opt/rustup/.rustup \
     CARGO_HOME=/opt/rustup/.cargo sh -c "\
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain nightly-2022-03-08 && \
-    rustup component add rust-src && \
-    rustup target add i686-unknown-linux-gnu && \
-    rustup target add riscv32imac-unknown-none-elf && \
-    rustup target add thumbv7em-none-eabihf && \
-    rustup target add thumbv7em-none-eabi && \
-    rustup target add thumbv7m-none-eabi && \
-    rustup target add thumbv6m-none-eabi && \
-    rustup target add thumbv8m.main-none-eabihf && \
-    rustup target add thumbv8m.main-none-eabi && \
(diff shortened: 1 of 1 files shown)


chrysn:
Removing src from beta makes no noticeable difference (still 11.4 / 2.6).

So, are these 300MB (1GB unpacked) something we want to impose on builders and users? If not, do we have alternatives?

(I've mentally started putting this into an executive-friendly why-or-why-not, but I'll belay that for when it is needed, maybe the decision is clear from what's around already).

kaspar030:
> If not, do we have alternatives?

an alternative would be a container that uses riot/riotbuild as base, just adding the rust stuff. then, in the build system, we implement some docker container selection based on modules used.

CI can handle the full container.

chrysn:
That does sound doable; I just wonder then where the right cutting point would be. Cutting between "nightly rust" and "beta rust" is a tad tricker because it would need to look into CARGO_CHANNEL, but if something like this were in place, pulling the with-rust image whenever the rust_target feature is used could happen.

I assume these modules will need to work linearly -- we can't have one for ESP and one for Rust and combine them all, for both might install something through apt and then there'd be a "merge conflict" on dpkg files.

kaspar030:
but, do we need both for PRs in master to go through? otherwise, why don't we stick to some nightly version, or just a specific beta?

chrysn:
(Clarifying: the stable is needed in the applicable tests to ensure it stays usable on stable, and the nightly is needed out of necessity for the rest).

kaspar030 (approved):
ACK.

Discussed in weekly, consensus not to block.

kaspar030:
bors merge

PR #195, opened 2022-05-13
Update C2Rust to 0.16

chrysn:
Things got a lot easier since upstream did awesome work leading up to https://github.com/immunant/c2rust/issues/365.

This release allows riot-sys to do away with a quite a few workarounds, and resolves a blocker for https://github.com/RIOT-OS/RIOT/pull/18056 in passing.

This is a prerequisite for https://github.com/RIOT-OS/riotdocker/pull/189 (in the course of which C2Rust can then be simplified further, as we can just use Rust from Ubuntu to build it, or even simply let this become a Debian package over time).

### DIFF
--- a/c2rust-built/Dockerfile
+++ b/c2rust-built/Dockerfile
@@ -68,6 +68,12 @@ RUN \
     apt-get -y install /c2rust_*.deb && \
 #     true
 #
+# # Provide the package in an easy to match place independent of version and
+# # architecture
+# RUN \
+    cp -l /c2rust_*.deb /c2rust.deb && \
+#     true
+#
 # # TBD: Deduplicate list with above; feisty apt-get has no --mark-auto,
 # # otherwise we could just --mark-auto above and autoremove --purge here.
 # RUN \
--- a/c2rust-built/Dockerfile
+++ b/c2rust-built/Dockerfile
@@ -10,7 +10,6 @@ COPY debian/control /dpkg-build/debian/
 # stay up there as it influences what will be installed.
 COPY debian/rules /dpkg-build/debian/
 COPY debian/changelog /dpkg-build/debian/
-COPY debian/install /dpkg-build/debian/
 
 # noninteractive for the tzinfo
 RUN \
@@ -25,7 +24,6 @@ RUN \
         devscripts \
         dpkg-dev \
         equivs \
-        git \
         && \
     mk-build-deps -i /dpkg-build/debian/control -t 'apt-get -y --no-install-recommends' && \
     echo 'Clean up installation files' >&2 && \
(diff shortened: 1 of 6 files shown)


chrysn:
bors try

(I'd like to see whether this builds fine with the currently pushed preliminary c2rust image; once this is approved but before it is m3rged, I'd re-tag that and remove the top commit that says "DO NOT M3RGE")

PS. I'm just spelling weirdly to not give b0rs any weird ideas on which keywords to match.

chrysn:
This is still running into trouble on RISC-V (which I could previously not test because its C side requires building in docker on my system), investigating... (Plus: Why didn't this show here during the "Rust build test" phase?)

chrysn:
On a side note, now that things work without an extra Rust toolchain, we could simplify things by just doing the c2rust build inside riotbuild. (This was originally impossible because we needed an odd nightly). There is still an upside to the current approach, that is, that the Debian package around it makes it easy to remove build dependencies later while keeping runtime dependencies.

If we dropped the current mechanism, we'd wind up with libclang-dev, libssl-dev and llvm-dev installed in the containers. According to apt, that comes at a cost of 56MB (downloads, that's probably also the compressed delta) / 424MB (uncompressed). They could be uninstalled again after building c2rust, but that'd need careful consideration of which runtime components need to be retained (which is what the Debian package does for us).

The PR works as it is now (also for RISC-V at the cost of using a git version rather than the released one again, but now that's easier); if the reviewers want to get rid of the complexity of the extra docker image, I can shift over to doing it in riotdocker right away.

kaspar030:
> if the reviewers want to get rid of the complexity of the extra docker image, I can shift over to doing it in riotdocker right away.

the reviewers might like that c2rust doesn't get rebuilt so much. :)

kaspar030 (approved):
ACK.

kaspar030:
bors merge

kaspar030:
did you consider PRing the debian package generation upstream?

kaspar030:
ok :)

bors merge

chrysn:
I had to force push once to get rid of the "DO NOT MERGE" commit that had previously ensured that the preview c2rust image was pushed; now that I pushed to the right tag (after approval of the PR) things can build as was before. Therefore,

bors merge

again.

> the reviewers might like that c2rust doesn't get rebuilt so much. :)

Trust me, so would I. That now it is running on stable removes a lot of volatility. Can't promise that this is the last for some time, due to the RISCV bug I missed in 0.16, but I do hope it's getting better.

If we had better sharing of built code between examples, we might even get rid of it as a binary completely (and have Cargo build it as build-dependency locally on demand).

> did you consider PRing the debian package generation upstream?

Yes -- but in the end the place where that should go is Debian directly.

PR #196, opened 2022-05-13
Simplify c2rust

chrysn:
This is the change hinted at in https://github.com/RIOT-OS/riotdocker/pull/195#issuecomment-1126075139 -- it does away with all the troublesome complexity (different images whose dependencies are intransparent to the CI builds).

On the downside, it increases the compressed image size by 56MB (estimated). The reduction in complexity should be well worth it. Compared to before (when a C2Rust update was an exercise of confirming a PR without merging it, then updating a docker tag manually, then letting the actual build through in a way that's not trivial to roll back), the whole thing is now just a block of dependencies and a single line of installation.

Also, this is sneaking in a fixup in our branch of C2Rust that allows unblocking https://github.com/RIOT-OS/RIOT/pull/18056 for good now.


chrysn:
Unless urgency comes to the blocked issue, I'm not pushing this actively until immunant push their next C2Rust release, as that'd allow further simplification. (The C2Rust version referenced in this PR is still a branch of mine; they already merged that, but I can't easily reference "the current master they have", and when they release it'll be shorter again).

Once that's through I should also update installation docs (they still refer to an older C2Rust version, which should still work though, as the workarounds are not removed yet).

chrysn:
I'd prefer to have #199 merged instead; this is only left open in case something unexpected pops up there. (I'm running *a* recent C2Rust version locally but not the exact one I'm pinning there, but then again the changes appear to be trivial)

chrysn:
Also, I have reason to believe that if this PR were started again, it'd fail with similar symptoms as #199; abandoning this in full.

PR #199, opened 2022-06-14
Simplify c2rust (with latest C2Rust version)

chrysn:
This is equivalent to #196 (also fixing https://github.com/RIOT-OS/RIOT/pull/18056) except that it is

* rebased already, and
* uses the latest upstream version after upstream merged all my relevant pull requests (they're still a bit slow on doing a Cargo release, but installing from a rev in their git history is actually easy).

(Opened as a separate PR because if this fails for some reason, it's easy to fall back to #196).


chrysn:
Build failure is being investigated, looks like they broke something in the Cargo.toml workspace setup -- it's a build system issue.

chrysn:
Downgraded C2Rust in a hotfix again by a bit: Starting [a commit later](https://github.com/immunant/c2rust/commit/0f5a0943b29b14c9162f13b9a9a63a08ab4a81f8) it is using a more recent Rust version that we currently ship (the latest stable) -- not a problem to upgrade, but not a thing I'd want to tangle in here either.

chrysn:
nagh ... and now *something* is leaving root-owned files in /data/riotbuild; debugging...

riotbuild/Dockerfile
```diff
@@ -309,6 +312,12 @@ RUN \
         true; \
     done"
 
+RUN \
+    echo 'Installing C2Rust' >&2 && \
+    CARGO_HOME=/opt/rustup/.cargo cargo install --no-track --locked c2rust --git https://github.com/immunant/c2rust --rev ed56c794230bbf5a0e417941e4778316d31ab114 && \
+    echo 'Cleaning up root-owned crates.io cache' >&2 && \
```
aabadie, line 318: Maybe set the rustup version in an ARG ?

```suggestion
ARG RUSTUP_VERSION=ed56c794230bbf5a0e417941e4778316d31ab114
RUN \
    echo 'Installing C2Rust' >&2 && \
    CARGO_HOME=/opt/rustup/.cargo cargo install --no-track --locked c2rust --git https://github.com/immunant/c2rust --rev $RUSTUP_VERSION && \
    echo 'Cleaning up root-owned crates.io cache' >&2 && \
```

chrysn: I expect the next change on this to be switching from a commit version to a cargo released version, which would then be a different line (no --git, no --rev) again. Then it'd make sense, now I see little value in it.

aabadie: Ok I see, let's leave it like this then!

kaspar030 (approved):
ACK.

kaspar030:
bors merge

PR #200, opened 2022-07-22
riotbuild: add ESP32-S3 toolchain

gschorcht:
Now that the 2022.07 release branch is created, it is a good time to go the next migration step and to add Espressif's vendor toolchain for ESP32-S3 to be able to compile RIOT-OS/RIOT#18185 in CI. The toolchain has a size of 292 MByte.

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -225,14 +225,23 @@ RUN echo 'Installing ESP32 QEMU' >&2 && \
     curl -L ${ESP32_QEMU_URL} | tar -C /opt/esp -xj
 ENV PATH $PATH:/opt/esp/qemu/bin
 
-# Install ESP32C3 toolchain in /opt/esp (344 MB)
+# Install ESP32-C3 toolchain in /opt/esp (344 MB)
 ARG ESP32_GCC_FILE=riscv32-esp-elf-${ESP32_GCC_VERSION}-${ESP32_GCC_RELEASE}-linux-amd64.tar.gz
 ARG ESP32_GCC_URL=${ESP32_GCC_REPO}/${ESP32_GCC_RELEASE}/${ESP32_GCC_FILE}
 
-RUN echo 'Installing ESP32C3 toolchain' >&2 && \
+RUN echo 'Installing ESP32-C3 toolchain' >&2 && \
     curl -L ${ESP32_GCC_URL} | tar -C /opt/esp -xz && \
     pip install --no-cache-dir pyserial
 ENV PATH $PATH:/opt/esp/riscv32-esp-elf/bin
+
+# Install ESP32-S3 toolchain in /opt/esp (292 MB)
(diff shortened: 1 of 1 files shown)


benpicco (approved):

benpicco:
Does ESP32-S2 work with the existing toolchain?

gschorcht:
> Does ESP32-S2 work with the existing toolchain?

Nope, that requires a separate toolchain that I could provide.

benpicco:
Please do, I actually have one of those boards :smile: 

I'm still surprised, aren't they all using the same Xtensa core? Why are separate toolchains needed for different members of the family?

gschorcht:
> > Does ESP32-S2 work with the existing toolchain?
> 
> Nope, that requires a separate toolchain that I could provide.

I could provide it immediately. The question is whether it makes sense before we have ESP32-C3 and ESP32-S3 support in the master. Each toolchain blows up the Docker image by about 300 MBytes.

benpicco:
As it can take very long for the workers to get updates, I say we better have toolchain support earlier than later.

gschorcht:
> Please do, I actually have one of those boards smile

Hm, but the ESP32-S2 support bases on some changes introduced for ESP32-S3 which in turn bases on some changes that can't be merged independent on the PR for ESP32-C3 support. That is, we have to merge the support for each variant step by step :worried:

> I'm still surprised, aren't they all using the same Xtensa core? Why are separate toolchains needed for different members of the family?

I don't know. At least ESP32 uses a different Xtensa core than ESP32-S2 and ESP32-S3. While ESP32 is using LX6 cores, new ESP32x variants like ESP32-S2 and ESP32-S3 use LX7 core.

gschorcht:
> I don't know. At least ESP32 uses a different Xtensa core than ESP32-S2 and ESP32-S3. While ESP32 is using LX6 cores, new ESP32x variants like ESP32-S2 and ESP32-S3 use LX7 core.

As I know, the cores differ in the instruction sets they implement. Xtensa cores are highly configurable by the SoC designer. There are a lot of hardware options from which a designer can select what he wants. As a result, supported instructions can differ a lot. I guess, the compiler has to know what instructions are supported so that it has to be built for each Xtensa core.

benpicco:
> the ESP32-S2 support bases on some changes introduced for ESP32-S3 which in turn bases on some changes that can't be merged independent on the PR for ESP32-C3 support. That is, we have to merge the support for each variant step by step 

Don't worry about that, only more motivation for me to review the C3 and S3 PRs :wink: 
I'll try to find some time for it on the weekend.

gschorcht:
Hm, the build test seem to fail due to an unrelated error in Rust build test. It seems that a `git checkout` command is used and the `examples/rust-hello-world/Cargo.lock` file is modified during the compilation of  `examples/rust-hello-world/`.
```
error: Your local changes to the following files would be overwritten by checkout:
	examples/rust-hello-world/Cargo.lock
Please commit your changes or stash them before you switch branches.
Aborting
Error: Process completed with exit code 1.
```

benpicco:
Bors try

benpicco:
Bors merge

gschorcht:
@benpicco Thanks for reviewing and merging.

PR #204, opened 2022-09-14
rust: Update nightly version

chrysn:
This allows more precise testing of changes that have stabilized. Concretely, all the changes (cstr in core, generic associated types) that have been stabilized available in the nightly in the form in which they will later be part of the releases.

Testing: CI should do (given it runs through representative builds), but I'm doing some extra tests (building out-of-tree examples) just to ensure that coverage is even better and make it even less likely to cause problems once a first full run is done on murdock.


chrysn:
Whoops, this will probably build-fail because a Rust library hasn't been updated in a while. I'm drilling down on it, probably just needs a RIOT-side round of updates.

chrysn:
My tests wound up positive, and CI is also happy.

kaspar030 (approved):
ACK.

kaspar030:
bors merge

PR #207, opened 2022-09-25
rust: Update nightly version

chrysn:
This is what https://github.com/RIOT-OS/riotdocker/pull/204 should have been, but as that nightly build did not contain that day's rustc, the intended version was not installed.

---

I did tests back in #204, but they were about "do they existing things still work", not (what I couldn't test easily back then, as it'd have spanned many crates at once) "does it also have the intended effect of allowing stable use of all the new features".

This unblocks https://github.com/RIOT-OS/rust-riot-wrappers/pull/14, and would have made the rollback of https://github.com/RIOT-OS/RIOT/pull/18642#issuecomment-1257159506 unnecessary.

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -325,9 +325,9 @@ ENV RUSTUP_HOME /opt/rustup/.rustup
 RUN \
     RUSTUP_HOME=/opt/rustup/.rustup \
     CARGO_HOME=/opt/rustup/.cargo sh -c "\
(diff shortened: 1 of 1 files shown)


chrysn:
This should be the last of the "update nightly to get X" updates.

After the 2022.10 release, I'd consider removing nightly from the images altogether; no modules in the ecosystem need nightlies any more, and when users build applications, they pretty please use a local Rust -- see https://github.com/RIOT-OS/riotdocker/issues/208.

kaspar030 (approved):
ACK.

kaspar030:
bors merge

PR #209, opened 2022-09-27
CI: Test Rust both with hello-world and gcoap example

chrysn:
Build failures after https://github.com/RIOT-OS/riotdocker/pull/189 show that the test coverage is insufficient to catch C2Rust breakage; this mitigates it by building an example that pulls in much more code.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -141,6 +141,7 @@ jobs:
           # action does only minimal fetching.
           (cd RIOT && git fetch origin master && git checkout FETCH_HEAD)
           make -CRIOT/examples/rust-hello-world BUILDTEST_MAKE_REDIRECT='' buildtest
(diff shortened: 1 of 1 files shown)


kaspar030 (approved):
ACK.

kaspar030:
this will currently fail, right?

bors merge

chrysn:
Yes, I hope so. (Otherwise the issue would be way harder to track).

I have no clue whether the follow-up #210 would pass.

chrysn:
Sadly, it did pass. Curious for bors' result, but AIU it's doing the very tests that the build target does (it just merges with master before, but as this forks off master, that's a no-op).

chrysn:
It turned out that this was *not* the issue (rather, no RISC-V have been run at all), but still, this gives *way* better code coverage, so it's still a win.

PR #210, opened 2022-09-27
C2Rust updates / fixes

chrysn:
This branch is used for investigating failures encountered after merging #189, in which C2Rust errs internally. Upstream might already have solved it.

This builds on #209.


chrysn:
Hmpf. It's not helping, but not breaking things either :-/

chrysn:
It's now not only updating to latest upstream, but to a branch that fixes the concrete issue upstream.

I'd leave this as two commits because these can well be understood separately, and that the first passed CI on its own (even though that was only because RISCV was untested at that time; now it should not pass any more thanks to https://github.com/RIOT-OS/RIOT/pull/18654) is good separate information.

kaspar030:
LGTM. Let's wait if the hifive1 rust code is built now

kaspar030:
bors merge

Issue #208, opened 2022-09-25
Decide future of nightly Rust in images


chrysn:
After the Rust 1.65 release (mid-November), we won't have an urgent need for nightly Rust any more. Options then are:

* Remove nightly Rust from our images.

  Users who build applications that do add nightly-only crates couldn't use our images, just like users who build applications that need anything else fancy in the build system are on their own.

* Unpin nightly, and run a test or two additionally with nightlies in the image build process.

  This would help users who use such crates, and would also help us spot regressions in stable Rust that hit us (I've never seen such a thing, but we'd contribute to keeping it that way) early on, whenever we update riotdocker.

  The nightlies would generally grow outdated at some point, but never be older than the stable Rust we ship, and would be easy to refresh by re-running the Action if we go without any other riotbuild updates for a long time.

I haven't made up my mind, don't have preferences yet, and appreciate input.

chrysn:
Right now I'm leaning toward just removing nightly; I don't see any upcoming components that'd need nightly again, and it's the easier path. There's nothing that keeps us from re-introducing it on demand.

In terms of keeping upcoming versions tested, we might consider adding a second container (possibly built on the regular one) that takes nightly and makes it the default toolchain. That container would be useful for those who build Rust code on RIOT and add crates that depend on nightly; it'd get the regular testing any container gets (without, obviously, any guarantee that it builds any nightly-dependent code that's not even in the RIOT tree).

PR #214, opened 2022-11-03
rust: Only install stable toolchain

chrysn:
Closes: https://github.com/RIOT-OS/riotdocker/issues/208

As of Rust 1.65, there is no need for nightly versions of Rust for RIOT any more.

Currently, this will likely (i.e. unless the nightly selection script happens to manage the case of no nightlies by assigning stable instead) not build, and depends on https://github.com/RIOT-OS/RIOT/pull/18839 which switches away from nightlies. (See there for an overview of all lockstepping involved)

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -342,21 +342,18 @@ ENV RUSTUP_HOME /opt/rustup/.rustup
 RUN \
     RUSTUP_HOME=/opt/rustup/.rustup \
     CARGO_HOME=/opt/rustup/.cargo sh -c "\
-    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain nightly-2022-09-25 && \
-    rustup toolchain add stable && \
-    for T in nightly-2022-09-25 stable; do \
-        rustup component add rust-src --toolchain \$T && \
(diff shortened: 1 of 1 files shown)


chrysn:
AIU backports don't use tagged CI builds, so this is stalled until the 2023.01 release makes 2022.10 unsupported.

(Unless we prefer to backport https://github.com/RIOT-OS/RIOT/pull/18839 -- that'd clear CI; release users, unlike CI, do use tagged dockers as far as I know. But my preference is to just let it grow out in 2023.01.)

chrysn:
With https://github.com/RIOT-OS/RIOT/pull/18839 and https://github.com/RIOT-OS/RIOT/pull/19058 merged, we can now do this.

To make sure this is tested well, I'd prefer to merge this after https://github.com/RIOT-OS/riotdocker/pull/182. Either way, even when ACKed, I'd leave it to the current RM to merge.

chrysn:
Now that all relevant changes are long in, *and* CI likes us again, I'd like to ask for final review on this.

(Also, the nightly version we had pinned is so ancient that it doesn't support features that riot-wrappers is starting to take for granted because they're stable already).

maribu (approved):

chrysn:
Seems we're still not really in hard freeze yet, so squishing it in:

bors merge

PR #215, opened 2022-11-04
c2rust: Use released version

chrysn:
This reduces the dependency on a particular GitHub branch that I was already about to remove after https://github.com/immunant/c2rust/pull/693 was done.

Still marked as a draft as it depends on https://github.com/immunant/c2rust/issues/690 (but that has all its blockers gone already).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,6 +133,12 @@ jobs:
           # that user, blocking downloads of own libraries.
           rm -rf ~/.cargo
           # Something left directories that break things
+          # A git clean wouldn't remove repositories
+          rm -rf RIOT/build
+          # These files can't be removed?
+          ls -ld tests/pkg_nanopb/bin/native/core_lib
+          ls -l tests/pkg_nanopb/bin/native/core_lib/atomic_c11.d
+          rm -rf tests/pkg_nanopb/bin/native/core_lib
           (cd RIOT && git clean -fxd)
           make -CRIOT/examples/rust-hello-world BUILDTEST_MAKE_REDIRECT='' buildtest
           make -CRIOT/examples/rust-gcoap BUILDTEST_MAKE_REDIRECT='' buildtest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -132,6 +132,8 @@ jobs:
           # Some of the above are executed by root, creating ~/.cargo/git as
           # that user, blocking downloads of own libraries.
           rm -rf ~/.cargo
+          # Something left directories that break things
+          (cd RIOT && git clean -fxd)
           make -CRIOT/examples/rust-hello-world BUILDTEST_MAKE_REDIRECT='' buildtest
           make -CRIOT/examples/rust-gcoap BUILDTEST_MAKE_REDIRECT='' buildtest
         env:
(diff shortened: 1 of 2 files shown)


kaspar030 (approved):
ACK.

chrysn:
Reopening after accidentally having deleted the branch.

chrysn:
Huh, saying that this is not where I'd have expected the error to show up would be putting it mildly -- I'm having a look at why this fails, but it may take a while.

chrysn:
I have no clue what caused the build failure, and can not reproduce the issue locally (after building the container in podman and using `BOARDS="arduino-uno esp32-wroom-32 hifive1b msb-430h native samr21-xpro" make -C examples/rust-gcoap BUILD_IN_DOCKER=1 DOCKER=podman DOCKER_IMAGE=localhost/riotbuild:latest buildtest` to run the very test that CI should have run).

Let's see if it was a random glitch (like, file system error?), restarting the build test.

chrysn:
Seems that builds break due to files left over from other runs. Switching to draft PR to prevent accidental merging just because a test happens to fix it.

chrysn:
I suspect now that this is due to changes in some base image, and would break in master just as well.

I still can't make progress for lack of a way to reproduce it (works locally), so I'm triggering a master rebuild to cut down the space.

.github/workflows/build.yml
```diff
@@ -132,6 +132,8 @@ jobs:
           # Some of the above are executed by root, creating ~/.cargo/git as
           # that user, blocking downloads of own libraries.
           rm -rf ~/.cargo
+          # Something left directories that break things
+          (cd RIOT && git clean -fxdv)
```
JKRhb, line 136: Looks like the CI error is caused by the `-v` flag here:

```suggestion
          (cd RIOT && git clean -fxd)
```

.github/workflows/build.yml
```diff
@@ -132,6 +132,14 @@ jobs:
...
           # that user, blocking downloads of own libraries.
           rm -rf ~/.cargo
+          # Something left directories that break things
+          # A git clean wouldn't remove repositories
+          rm -rf RIOT/build
+          # These files can't be removed?
+          ls -ld tests/pkg_nanopb/bin/native/core_lib
+          ls -l tests/pkg_nanopb/bin/native/core_lib/atomic_c11.d
+          rm -rf tests/pkg_nanopb/bin/native/core_lib
+          (cd RIOT && git clean -fxd)
```
crasbe, line 142: This got addressed by #290 and https://github.com/RIOT-OS/RIOT/pull/22631

riotbuild/Dockerfile
```diff
@@ -344,7 +344,7 @@ RUN \
 
 RUN \
     echo 'Installing C2Rust' >&2 && \
-    CARGO_HOME=/opt/rustup/.cargo cargo install --no-track --locked c2rust --git https://github.com/chrysn-pull-requests/c2rust --branch riscv-vector-types && \
+    CARGO_HOME=/opt/rustup/.cargo cargo install --no-track --locked c2rust --version 0.17.0 && \
```
crasbe, line 347: And somewhere along the way this was already done, with #283 we're at 0.22.1 anyways.

crasbe:
Everything has been addressed by other PRs, so this can be closed.

PR #216, opened 2022-11-16
murdock_worker: add newer ccache (4.7.4)

kaspar030:
ccache version 4.7 adds the option to have a remote-only cache on secondary storage (e.g., **redis**).

I have a [branch](https://github.com/RIOT-OS/murdock-worker/pull/5) modifying the murdock worker docker-compose to add a redis instance for caching.

The ccache binary is the official binary release from https://ccache.dev/download.html.
(debians bookworm has a .dpkg, but that's not compatible to our Ubuntu's older libraries).

My hope is that a local redis scales better with larger caches, which have shown to stall builds when the cache is full and needs to be cleaned (@cgundogan remember the sawtoothing?). redis LRU is cheap.

Also, redis handles persistence (compared to tmpfs), which would **make reboots/restarts not loose all of ccache**.

Potentially, redis can be distributed, so e.g., the mobis could have a larger cache shared between them.

(I've piggibacked a commit creating `/cache` with rwx:all permissions, that's necessary to mount volumes so they're accessible by a non-root in-docker uid. otherwise, mounts default to root-only permissions.)

### DIFF
--- a/murdock-worker/.gitattributes
+++ b/murdock-worker/.gitattributes
@@ -1,0 +1,1 @@
+*.tar.xz filter=lfs diff=lfs merge=lfs -text
--- a/murdock-worker/Dockerfile
+++ b/murdock-worker/Dockerfile
@@ -42,6 +42,9 @@ RUN tar -xvf /${CCACHE_TGZ} --strip-components=1 -C/usr/local/bin && rm /${CCACH
 # install murdock slave startup script
 COPY murdock_slave.sh /usr/bin/murdock_slave
 
+# create cache folder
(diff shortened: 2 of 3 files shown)


kaspar030:
- bumped to 4.7.4
- needs squashing now

kaspar030:
I think we should go with this. It would make testing redis as cache backend much easier. And rule out potential performance oddities when comparing old ccache on tmpfs vs new ccache on redis.

maribu (approved):

kaspar030:
bors merge

PR #218, opened 2022-12-16
Dockerfile: install libsdc++-arm-none-eabi-*

maribu:
As the title says...

Should fix https://ci.riot-os.org/results/8e1338abe27b407c8d2677e3a42ebb7e/output/builds/examples/riot_and_cpp/nrf52840dk:gnu.txt and similar failures.

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -67,6 +67,10 @@ RUN \
         cython3 \
         gcc-multilib \
         gcc-arm-none-eabi \
+        libnewlib-arm-none-eabi \
+        libstdc++-arm-none-eabi-dev \
+        libstdc++-arm-none-eabi-newlib \
+        libstdc++-arm-none-eabi-picolibc \
         gcc-riscv64-unknown-elf \
         gcc-xtensa-lx106 \
         gdb \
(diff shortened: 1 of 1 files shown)


aabadie (approved):

aabadie:
bors try

maribu:
I have to admit that I fail to understand Ubuntu's package management here

- Why are things inconsistent between RISC-V and embedded ARM toolchains? (E.g. I assumed that I would have to install the same `libstdc++-riscv64-unknown-elf-*` packages, but they do not exists.)
- Why is there a newlib and a picolibc version of libstdc++ - even though neither actually implements libc++? (In fact, libstdc++ is provided by g++ / GCC and should be the exact some no matter if newlib or picolibc is used)
- Why is there a -dev version? Nobody will need to have static libs for an embedded ARM toolchain arround (that do not link with any local executables) unless one does embedded development, for which one needs both headers and static libs. It makes IMO zero sense to split them.

aabadie:
I opened #219 so the CI also checks c++ applications can also be built without errors. I'm not sure if the list of board is enough (but it seems so)

aabadie:
bors merge

PR #221, opened 2023-01-17
actions: remove obsolete switcharoo

chrysn:
Copy of https://github.com/RIOT-OS/riotdocker/pull/182 that is rebased (actually, the logical removals are re-applied, but it's dressed with the same author and commit message).


kaspar030 (approved):
ACK.

kaspar030:
bors merge

PR #226, opened 2023-02-25
[WIP] riotbuild: Split out build logic

maribu:
This is work in progress and untested yet


maribu:
I did some basic testing. The docker image is a bit smaller (10.7 GB vs 14.1 GB) and the new RISC-V toolchain generates smaller binaries. I think the new picolibc version is what was left to switch to picolibc by default. My limited test set showed no compilation failure for a few boards and tests. I think this is promising.

PR #227, opened 2023-04-04
riotbuild: bump esp32x toolchains to gcc version 12.2

gschorcht:
This PR upgrades ESP32x toolchains to GCC version 12.2. Since ESP-IDF 5.1 requires this compiler version, this PR has to be merged before we can start with RIOT upgrade to ESP-IDF 5.1.

This PR depends on https://github.com/RIOT-OS/RIOT/pull/19450.

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -204,19 +204,19 @@ RUN echo 'Installing ESP8266 toolchain' >&2 && \
 ENV PATH $PATH:/opt/esp/xtensa-esp8266-elf/bin
 ENV ESP8266_RTOS_SDK_DIR /opt/esp/ESP8266_RTOS_SDK
 
-# Install ESP32 toolchain in /opt/esp (289 MB)
-ARG ESP32_GCC_RELEASE="esp-2021r2-patch3"
-ARG ESP32_GCC_VERSION="gcc8_4_0"
+# Install ESP32 toolchain in /opt/esp (338 MB)
+ARG ESP32_GCC_RELEASE="esp-12.2.0_20230208"
+ARG ESP32_GCC_VERSION="12.2.0_20230208"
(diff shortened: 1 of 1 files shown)


kaspar030:
> @kaspar030 The build test had to fail when I pushed the PR because [RIOT-OS/RIOT#19450](https://github.com/RIOT-OS/RIOT/pull/19450) was not merged. Now it should succeed. How can we restart the build test?

I guess someone restarted it?

Anyhow, it probably still won't succeed. This repository doesn't build RIOT master but some older branch as defined [here](https://github.com/RIOT-OS/riotdocker/blob/c26aa61b4c0ef2af3386713d9719364c6ae364b7/.github/workflows/build.yml#L38-L39).

That needs to be bumped first. (Which we usually rather not do during final release testing ...)

Can this wait until after the release?

kaspar030:
> After the release we should not wait too long before inconsistencies happen between the last state of ESP32 code and the new compiler version.

I PR'ed the [version bump](https://github.com/RIOT-OS/riotdocker/pull/228), we can rerun this.

kaspar030:
Pls rebase!

gschorcht:
Rebased, but it seems that the compile test still used the old code base. When you look for example on [the first error in the log](https://github.com/RIOT-OS/riotdocker/actions/runs/4839586727/jobs/8624699619?pr=227#step:9:33), this problem was changed with [PR 19450](https://github.com/RIOT-OS/RIOT/blob/bf95fa49092e3fab103ee658cac0ebd6cd2c98ca/cpu/esp32/periph/gpio.c#L232).

kaspar030:
- triggered a re-run

kaspar030 (approved):
ACK.

kaspar030:
bors merge

PR #234, opened 2023-10-12
Pin down scipy version

chrysn:
Newer versions cause build failures because they expect a newer cython than is provided in jammy; they'd show errors like:

```
      Error compiling Cython file:
      ------------------------------------------------------------
      ...
      ctypedef float s                                                                                                                             ctypedef double d                                                                                                                            ctypedef float complex c
      ctypedef double complex z
      cdef void caxpy(int *n, c *ca, c *cx, int *incx, c *cy, int *incy) noexcept nogil
                                                                        ^
      ------------------------------------------------------------

      /usr/local/lib/python3.10/dist-packages/scipy/linalg/cython_blas.pxd:20:67: Syntax error in C variable declaration
```

Let's hope this allows building again -- if not, we have multiple bitrots to combat.

### DIFF
--- a/riotbuild/requirements.txt
+++ b/riotbuild/requirements.txt
@@ -11,6 +11,7 @@ colorama>=0.4.0
 cryptography==3.3.2
 scapy>=2.4.3
 protobuf==3.18.3
+scipy == 1.10.1            # needed by scikit-learn, but newer ones fail with jammy's cython
 scikit-learn==0.22.1
 joblib==1.2.0
 emlearn==0.10.1


chrysn:
For context, I also tested 1.9.0 (installation worked); at the time of filing this, 1.11.3 is current, which broke in the way described above. I didn't easily find when scipy changed its cython requirements, but it'd stand to versioning reason they'd only do that as part of at least the minor 1.11 release, and 1.10.1 is currently the last of the 1.10s. (Could probably do this more sharply as `< 1.11`, but we've been burnt enough here already, and I don't think that there will be security updates for scipy 1.10).

maribu (approved):

chrysn:
Thanks. We're not even in hard freeze, so I think I can just go:

bors merge

PR #235, opened 2023-12-12
ci: enable "build" for merge groups

kaspar030:
public bors has shut down. switch to merge_queues.

This adds another (third!) build run for each change. But it does keep the semantic conflicts out. So, let's get back the ability to merge and (maybe) optimize later.


kaspar030:
I'll fix up branch requirements as we go here, don't click "merge" yet.

kaspar030:
... disabling rust-gcoap for now ...

PR #244, opened 2024-02-06
murdock-worker: bump git-cache-rs to 0.1.5

kaspar030:
When the initial `git cache clone` would have a commit specified (`--commit HASH`), and would trigger the initial clone into the cache directory, the final "checkout HASH" was skipped.

This caused the first dwq worker thread to checkout a branch to fail every job with `./.murdock: not found`.

https://github.com/kaspar030/git-cache-rs/compare/0.1.3...0.1.5

### DIFF
--- a/murdock-worker/Dockerfile
+++ b/murdock-worker/Dockerfile
@@ -30,7 +30,7 @@ RUN pip3 install hiredis
 RUN pip3 install click
 
 # get git-cache-rs binary
-COPY --from=ghcr.io/kaspar030/git-cache:0.1.3-jammy /git-cache /usr/bin/git-cache
+COPY --from=ghcr.io/kaspar030/git-cache:0.1.5-jammy /git-cache /usr/bin/git-cache
 ENV GIT_CACHE_RS /usr/bin/git-cache
 
 # install newer ccache package


chrysn:
Please rebase onto https://github.com/RIOT-OS/riotdocker/pull/246 (or master once that is in), that should make the problem go away.

kaspar030:
huh, this was rebased, it still shows `fatal error: 'preprocessor_successor.h' file not found`?

chrysn:
/me facepalms

The 2024.01 release did not yet get the riot-sys fix.

On the "how do we make sure this doesn't happen again" side, I've added a check point to the [release check list](https://github.com/RIOT-OS/RIOT/wiki/Managing-a-Release) that'll flag when there are pending commits on riot-sys/-wrappers.

On the "what do we do now" side ... can we backport 002cd42e26e7fbc3140e9d7aa5a1788bfa9f70f9? Even if we don't do a point release (we rarely do), this makes sure the fix winds up on the branch that gets pulled. (I'd PR it myself, but the backport script doesn't work for people whose RIOT clone is not in their user repo).

kaspar030:
> I'd PR it myself, but the backport script doesn't work for people whose RIOT clone is not in their user repo

so which PR is that?

chrysn:
https://github.com/RIOT-OS/RIOT/pull/20303

MrKevinWeiss:
Or should it just be the one commit?

MrKevinWeiss:
> Even if we don't do a point release (we rarely do), this makes sure the fix winds up on the branch that gets pulled. (I'd PR it myself, but the backport script doesn't work for people whose RIOT clone is not in their user repo).

I would probably do a point release, maybe after a week if everything is stable.

kaspar030:
While you're here - would you give this an ACK? It's been tested quite a bit on ci-staging.riot-os.org.

chrysn (approved):

PR #247, opened 2024-02-12
static-test-tools: add PyYAML to requirements.txt

maribu:
This should allow parsing YAML files from Python as part of `static-tests` in the CI.


kaspar030 (approved):
ACK.

PR #248, opened 2024-02-12
riotbuild: update RISC-V Toolchain 

maribu:
Citing https://github.com/xpack-dev-tools/riscv-none-embed-gcc-xpack:

> DEPRECATED! Replaced by xpack-dev-tools/riscv-none-elf-gcc-xpack

So let's do that.

Note: Depends on and includes:

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -165,20 +165,20 @@ ENV MIPS_ELF_ROOT /opt/mips-mti-elf/${MIPS_VERSION}
 ENV PATH ${PATH}:${MIPS_ELF_ROOT}/bin
 
 # Install RISC-V binary toolchain
-ARG RISCV_VERSION=10.2.0-1.2
+ARG RISCV_VERSION=13.2.0-2
 RUN mkdir -p /opt && \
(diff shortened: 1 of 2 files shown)


kaspar030:
So, if we merge this, it might fail some builds in RIOT master and possibly the release branch. @maribu if you're "on it" I'm ok with merging this now.

kaspar030 (approved):
ACK. (merging when @maribu or others promise to watch RIOT master for fall-out)

maribu:
I think I can do some local testing first, e.g. all apps for a single RISC-V board should be doable on my machine.

maribu:
Yeah, there are some issue uncovered by the toolchain update. I think I can prepare a PR soonish that would fix that.

maribu:
This should be updated in lock-step with https://github.com/RIOT-OS/RIOT/pull/20380

kaspar030:
Let's go.

kaspar030:
So once this shows "merged", the master workflow needs to complete building & publishing the containers, then ~15min more to deploy, then we need to merge https://github.com/RIOT-OS/RIOT/pull/20380.

PR #251, opened 2024-03-21
Revert "ci: temporarily disable "Rust build test""

chrysn:
This reverts commit 96287a87b6118d0662bd5a822820d98a8af1261d from [249] that had been required to resolve a deadlock.

[249]: https://github.com/RIOT-OS/riotdocker/pull/249

---

Casually looking at the finished Rust update action I found that the "temporary" #249 has been around for over a month now. I don't remember what the deadlock was with, and none of us left a note, I presume it was something with #249 -- but at any rate, should be resolved now (and build will confirm that).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -128,22 +128,21 @@ jobs:
           DOCKER_IMAGE: ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
           BOARDS: "native samr21-xpro"
 
-            # TEMPORARILY disable rust test
-            #      - name: Rust build test
-            #        run: |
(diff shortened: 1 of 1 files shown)


maribu:
Wait, that arrow *is* green! I do have Unlimited power!

chrysn:
Our checkmarks have power, but merge pull requests is apparently still limited.

chrysn:
Also, there's no hurry to merge -- this can just as well merge it right before merging the next PR, this shouldn't have changed the build output, but our builds are not exactly binary reproducible.

kaspar030:
> Our checkmarks have power, but merge pull requests is apparently still limited.

hm, all Maintainers have "Write" permission, that should suffice actually to merge?

PR #252, opened 2024-09-09
Use pre-release version of c2rust 0.19

chrysn:
C2Rust has been troubled by the Rust 1.80 release, but is preparing a 0.19 release that'll give us a new stable base.

That version was announced in https://github.com/immunant/c2rust/pull/1116 but has not been uploaded; doing a bit of CI testing to verify it all works for us.

Before being ready, this PR will switch to using the released 0.19 version.


chrysn:
As discussed in https://github.com/immunant/c2rust/pull/1116, the Cargo release is delayed -- but as there is a git tag on it, we can install from there just as well. While that is not technically as irrevocable as Cargo release, git tags and GitHub releases are usually things that are not changed lightly.

Marking as ready-for-review, I think we can go with this.

benpicco (approved):

PR #253, opened 2024-10-02
static: Add cargo

chrysn:
This is needed for the static tests introduced in [20886]; no need for nightly or target specific tools.

[20886]: https://github.com/RIOT-OS/RIOT/pull/20886

This increases the size reported by `docker image list` from 781MB to 1.17GB.

### DIFF
--- a/static-test-tools/Dockerfile
+++ b/static-test-tools/Dockerfile
@@ -13,6 +13,7 @@ RUN \
     apt-get update && \
     echo 'Installing static test tools' >&2 && \
     apt-get -y --no-install-recommends install \
+        cargo \
         coccinelle \
         cppcheck \
         doxygen \


static-test-tools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG DOCKER_REGISTRY="docker.io/riot"
+arG DOCKER_REGISTRY="docker.io/riot"
```
Teufelchen1, line 1: What? Why?

Teufelchen1 (changes requested):

static-test-tools/Dockerfile
chrysn: No: How?

Honestly, no clue -- sorry it slipped in, thanks for spotting.

chrysn: (My best guess is that since my keyboard has ESC and ~ on the same key with different modifiers and ESC is something I use a lot, I hit ~ and doing so changed the case inadvertently.)

Teufelchen1 (approved):

chrysn:
My approach was not good: I missed that the static tools go into the container, and thus we all of a sudden have two Rust (one via Debian, one via rustup), and they don't like that.

I'll abandon this PR in favor of one that moves the Rust installation over from the docker image into the static tools.

PR #257, opened 2025-02-21
ci: bump RIOT_BRANCH and VERSION_TAG to 2025.01-branch and 2025.04

chrysn:
Regular update like https://github.com/RIOT-OS/riotdocker/pull/236 -- with this, we can start building containers that don't claim to be building for the older versions any more, which is one step in preparing for future releases (eg. where with https://github.com/RIOT-OS/riotdocker/pull/256 there will be changed checkout folders).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,8 +36,8 @@ jobs:
     name: Build and Test
     runs-on: ubuntu-latest
     env:
-      RIOT_BRANCH: '2024.01-branch'
-      VERSION_TAG: '2024.01'
+      RIOT_BRANCH: '2025.01-branch'
+      VERSION_TAG: '2025.04'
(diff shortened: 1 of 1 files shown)


chrysn:
I think we're on our own here; let's follow the pattern and learn as we go.

.github/workflows/build.yml
```diff
@@ -36,8 +36,8 @@ jobs:
     name: Build and Test
     runs-on: ubuntu-latest
     env:
-      RIOT_BRANCH: '2024.01-branch'
-      VERSION_TAG: '2024.01'
+      RIOT_BRANCH: '2025.01-branch'
+      VERSION_TAG: '2025.04'
```
benpicco, line 40: Does that tag exist yet?

mguetschow: Good point, probably it didn't back then o.O

PR #258, opened 2025-03-13
requirements.txt: Fix codespell version

MrKevinWeiss:
Master just started failing due to an update... Probably better to just fix it and update every once in a while after fixes are applied in RIOT master.


crasbe:
Now that I've (mostly) went through the upgrade process for Ubuntu 26.04 LTS, I'm not in favor of pinning the version.

I'd rather have stuff break along the way rather than having to do a bunch of things at once. While it would be nice to bump versions once in a while, the reality is that nobody really does it unless there's a reason to.

PR #264, opened 2026-01-12
riotbuild: upgrade ESP32 toolchain to GCC 15.2.0, GDB 16.3 and QEMU 9.2.2

gschorcht:


riotbuild/Dockerfile
```diff
@@ -206,9 +206,9 @@ ENV PATH $PATH:/opt/esp/xtensa-esp8266-elf/bin
 ENV ESP8266_RTOS_SDK_DIR /opt/esp/ESP8266_RTOS_SDK
 
 # Install ESP32x Xtensa toolchain in /opt/esp (1.1 GB)
```
crasbe, line 208: Are the installation sizes still accurate with the new versions?

gschorcht: I adjusted the sizes a little in commit ee53f51.

PR #268, opened 2026-05-06
riotdocker/requirements.txt: Reorder, group and document

crasbe:
The `requirements.txt` file has become a bit of a mess, so I dug through history and tried to find out which dependency is used for which purpose and group them accordingly.

Also, I bumped the version of `cryptography` to close #232 and a bunch of security warnings by @dependabot.

The build and test for `tests/sys/suit_manifest` still suceeds and considering that this was after a version jump from 3.3.2 to 48.0.0, I dared to un-pin the version of `cryptography` to avoid a future flood of @dependabot warnings.

I guess we'll see how that'll turn out 🤷‍♂️

### DIFF
--- a/riotbuild/requirements.txt
+++ b/riotbuild/requirements.txt
@@ -1,20 +1,36 @@
 # Required python libraries
-pyasn1==0.4.2
+# =========================
+#
+# Please try to maintain the alphabetical order of the entries within a
+# category and document their purpose (i.e. the module they're used for).
+
+# RIOT toolchain dependencies
+jinja2==2.11.3              # used by STM32 vector and IRQ file generation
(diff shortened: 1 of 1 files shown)


riotbuild/requirements.txt
```diff
@@ -1,20 +1,36 @@
 # Required python libraries
-pyasn1==0.4.2
+# =========================
+#
+# Please try to maintain the alphabetical order of the entries and document
```
mguetschow:
```suggestion
# Please try to maintain the alphabetical order of the entries within a category and document
```

mguetschow (approved):
Thanks for this work, LGTM! :)

crasbe:
Thanks :)

PR #276, opened 2026-05-11
.github/workflows: Check Secrets and update actions for NodeJS 24 Compilance

crasbe:
## Contribution

Similar to what I added in https://github.com/RIOT-OS/RIOT/pull/22085, the `riotdocker` workflows should also check if the Docker Hub credentials are set and if not, they shouldn't try to push images.

Before:

Screenshot 2026-05-11 154054

After:

Screenshot 2026-05-11 160402

While testing this, I noticed a slight warning, that NodeJS 20 based actions will be forced to run on NodeJS 24 by June 2026 and by September 2026, they will be disabled. See: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

## Testing

I tested it in my fork's `master` branch and it works:
image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,17 +61,17 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@main
         with:
           lfs: 'true'
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
+        uses: docker/setup-buildx-action@v4
         with:
           driver: docker
 
       - name: Build riotdocker-base
-        uses: docker/build-push-action@v2
+        uses: docker/build-push-action@v7
         with:
           context: ./riotdocker-base
           tags: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -33,35 +33,53 @@ on:
...
+    needs: [check-secret]
     env:
       RIOT_BRANCH: '2026.04-branch'
       VERSION_TAG: '2026.07'
       DOCKER_REGISTRY: "${{ secrets.DOCKER_REGISTRY || 'local' }}"
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@main
```
maribu, line 64: This doesn't look like the kind of action that one would expect breaking changes for. But better safe than sorry?

maribu (approved):

.github/workflows/build.yml
crasbe: In the main RIOT repo we use `actions/checkout@main` a lot, so I went with it here too.

PR #285, opened 2026-08-14
{riotbuild,static-test-tools}/Dockerfile: remove old dependencies

crasbe:
These have been present since the initial commit 12 years ago:
image

I have tried to find where they are used or if they are used at all, but they don't seem to be anymore.
`pcregep` was a requirement for the ESP8266 toolchain, but I don't think it is anymore (as it happily builds without):
image

### DIFF
--- a/riotbuild/Dockerfile
+++ b/riotbuild/Dockerfile
@@ -51,7 +51,6 @@ RUN \
         gdb \
         g++-multilib \
         libffi-dev \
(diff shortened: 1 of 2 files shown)


AnnsAnns (approved):

crasbe:
Found it: `Required command 'pcregrep' for './dist/tools/licenses/check.sh' not found in PATH`

PR #289, opened 2026-08-21
riotbuild/requirements.txt: bump protobuf version for nanopb

crasbe:
The dependency resolution in the Python version used by Resolute does not seem to include all the necessary dependencies anymore, so the build fails. This is also a good opportunity to update the version in general and close some security advisories.

Replaces #274.


crasbe:
There seems to be a bad constellation between the versions of Python, `protobuf` and `nanopb`. With Resolute it works, therefore I'll just include this in the Resolute update.

PR #290, opened 2026-08-27
.github/workflows: fix example paths in build workflows

crasbe:
It seems like the paths of the applications are incorrect and therefore they are not actually built. Note the `.../RIOT/RIOT/...` in the path.

image

Screenshot 2026-08-27 094651

Also the toolchain for the GNU build test was wrong (LLVM instead of GNU).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -120,13 +120,13 @@ jobs:
         env:
           BUILD_IN_DOCKER: 1
           DOCKER_IMAGE: ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
-          APPLICATIONS: RIOT/examples/basic/hello-world
-          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native samr21-xpro"
-          TOOLCHAIN: llvm
+          APPLICATIONS: examples/basic/hello-world
+          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native32 native64 samr21-xpro"
+          TOOLCHAIN: gnu
 
       - name: GNU microbit qemu test
         run: >
-          docker run --rm -t -v $(pwd)/RIOT:/data/riotbuild
+          docker run --rm --tty --user $(id -u):$(id -g) -v $(pwd)/RIOT:/data/riotbuild
           ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
           make -Ctests/sys/fmt_print all test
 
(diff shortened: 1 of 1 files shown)


crasbe:
I think this is a regression introduced in #256.

mguetschow (approved):
Ah fun, thanks for spotting! The failing test build seems unrelated and should probably be fixed outside of this PR.

crasbe:
> Ah fun, thanks for spotting! The failing test build seems unrelated and should probably be fixed outside of this PR.

Yes, it's an issue with our build system, especially the `tests-input-hash` target is not built in Docker and `BINDIR` is not passed to the Docker environment. I'll open a separate PR for that.

.github/workflows/build.yml
```diff
@@ -120,13 +120,13 @@ jobs:
...
-          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native samr21-xpro"
-          TOOLCHAIN: llvm
+          APPLICATIONS: examples/basic/hello-world
+          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native32 native64 samr21-xpro"
+          TOOLCHAIN: gnu
 
       - name: GNU microbit qemu test
         run: >
-          docker run --rm -t -v $(pwd)/RIOT:/data/riotbuild
+          docker run --rm --tty --user $(id -u):$(id -g) -v $(pwd)/RIOT:/data/riotbuild
```
crasbe, line 129: Setting the user explicitly was necessary because otherwise the files will be created as `root` and when the next command tries to delete the `build` directory, it fails due to missin permissions.

.github/workflows/build.yml
```diff
@@ -120,13 +120,13 @@ jobs:
         env:
           BUILD_IN_DOCKER: 1
           DOCKER_IMAGE: ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
-          APPLICATIONS: RIOT/examples/basic/hello-world
-          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native samr21-xpro"
-          TOOLCHAIN: llvm
+          APPLICATIONS: examples/basic/hello-world
+          BOARDS: "arduino-uno esp32-wroom-32 hifive1b msb-430h native32 native64 samr21-xpro"
```
crasbe, line 124: The `compile_test` script does not resolve board aliases, as it just compares the paramter with the result of `info-boards-supported`.

Therefore we have to set `native32` and `native64` explicitly.

.github/workflows/build.yml
```diff
@@ -159,12 +159,11 @@ jobs:
         env:
           BUILD_IN_DOCKER: 1
           DOCKER_IMAGE: ${{ env.DOCKER_REGISTRY }}/riotbuild:latest
-          # TODO: rust-gcoap temporarily disabled (sock_udp.h not found)
```
crasbe, line 162: Seems to build again. Temporary is over.

crasbe:
Thank you for the review :)

PR #286, opened 2026-08-14
treewide: update to Ubuntu 26.04 LTS Resolute Raccoon 🦝

crasbe:
This PR is currently WIP as it combines the previous PRs #280, #281, #282, #283, #284 and #285 and has to be rebased as they are merged.

This is the replacement for the previous effort in #239.

PR #292, opened 2026-09-10
riotbuild/Dockerfile: set `laze` and `git-cache` version tags to `resolute`

crasbe:
This is a follow-up for #286 once https://github.com/kaspar030/git-cache-rs/pull/144 and https://github.com/kaspar030/laze/pull/902 are merged.

This will pin the versions of `laze` and `git-cache` to `resolute` instead of `ubuntu latest`, which will change with every new Ubuntu release.


crasbe:
This is resolved as the PRs are already merged.

PR #295, opened 2026-09-12
{riotbuild,static-test-tools}/Dockerfile: move clang-format

crasbe:
To make sure the LLVM versions don't diverge, an additional check is added to make sure `clang-format` installed in static-test-tools has the same version as the other LLVM packages installed in riotbuild.

For https://github.com/RIOT-OS/RIOT/pull/22679


AnnsAnns (approved):

AnnsAnns:
🙏 ty

